PR #13, opened 2019-05-18
Server client pattern and Mediocre collision handling

SourenP:


src/Client.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+    public:
+        Client(int player_count);
+        ~Client();
+
+        void draw(sf::RenderTarget& target);
+        void update(std::vector<Group*> groups);
+
+    private:
+        std::vector<GroupShape*> mGroupShapes;
```
copacetic: The state in this class is a bit weird to me. Why does it keep group shapes but not groups? What is client meant to represent?

SourenP: Discussed offline

PR #19, opened 2019-06-02
Sync state over network

SourenP:
With this change you should be able to 
- See collision with two clients
- Quit on one client and see the circle go away on the other


src/server/main.cpp
```diff
@@ -9,7 +9,8 @@ int main(int, char const**) {
     //size_t player_1_id = game_controller.createPlayer();
 
     while (true) {
-        // Update game state based on client inputs
-        // game_controller.update(client_1.getId(), client_1.getDirection());
+        // TODO: What should this amount be set to?
+        sf::sleep(sf::milliseconds(50));
```
SourenP: Not sure what this amount should be. 
If I don't have this here then the server quickly moves the circles (as fast as it can compute) and the client doesn't have time to catch up, so the circle teleports super far away on the first _client_ state update.

This basically determines the games pace -  tick/sec

src/server/NetworkingServer.cpp
```diff
@@ -163,36 +180,53 @@ void NetworkingServer::ComputeGameState() {
...
-        for (const auto iter : mClientPositions) {
-            std::cout << iter.first << " " << iter.second[0] << " " << iter.second[1] << std::endl;
-        }
         // window is closed, time to reset. May want to use fix sized array as an optimization.
         // It's a bit mean to disguard old commands, maybe there should be windows for multiple ticks?
-        mClientMoves.clear();
         mCurrTick++;
+        // This sleep is needed to give the GameController time to fetch the client directions in getClientDirections
+        // TODO: Not sure what this sleep value should be...
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
SourenP: Not sure what this value should be. This is the window during which the server gets to read the move commands (mAcceptingMoveCommands == false). See getNetworkPlayers.

If I don't have a sleep here then mAcceptingMoveCommands is almost always true and the server doesn't get a chance to getNetworkPlayers and update player directions according to the client input.

src/client/NetworkingClient.cpp
```diff
@@ -44,16 +45,16 @@ sf::Uint32 NetworkingClient::Start() {
...
 
-std::vector<position> NetworkingClient::getPositions() {
-    while (!mAcceptingPositionRead) {
-        //
+std::vector<network_game_object> NetworkingClient::getNetworkGameObjects() {
+    if (mAcceptingNetworkGameObjectsRead) {
+        return mNetworkGameObjects;
     }
-    return mPositions;
+    return std::vector<network_game_object> {};
```
copacetic: why is this better?

src/client/NetworkingClient.cpp
```diff
@@ -44,16 +45,16 @@ sf::Uint32 NetworkingClient::Start() {
...
 }
 
-void NetworkingClient::setDirection(direction dir) {
+void NetworkingClient::setDirection(sf::Vector2f direction) {
```
copacetic: why change away from a dedicated struct for direction objects? what if we want to make direction 3d one day? don't really see the benefit here except maybe clarity

src/client/NetworkingClient.cpp
```diff
@@ -95,27 +97,25 @@ void NetworkingClient::RealtimeClientRecv() {
...
         if (packet >> server_tick) {
-            mAcceptingPositionRead = false;
-            mPositions.clear();
-            while (packet >> client_id >> x_pos >> y_pos) {
-                std::cout << "Client: " << client_id << " has position x,y: " << x_pos << " " << y_pos << std::endl;
-                position p = {client_id, x_pos, y_pos};
-                mPositions.push_back(p);
+            mAcceptingNetworkGameObjectsRead = false;
+            mNetworkGameObjects.clear();
+            while (packet >> client_id >> x_pos >> y_pos >> size) {
```
copacetic: shouldn't this be `packet >> ngo`?

src/client/NetworkingClient.cpp
```diff
@@ -95,27 +97,25 @@ void NetworkingClient::RealtimeClientRecv() {
...
+                network_game_object ngo = {client_id, x_pos, y_pos, size};
+                mNetworkGameObjects.push_back(ngo);
             }
             // Im not sure what kind of synchronization needs to happen here.
             // If this tick is the most up-to-date we've ever seen, maybe we set the game to it?
-            std::cout << "Client positions from server for tick: " << server_tick << std::endl;
             mCurrentTick = server_tick;
-            mAcceptingPositionRead = true;
+            mAcceptingNetworkGameObjectsRead = true;
```
copacetic: I wonder if it's important that the word `network` be in the name of this variable as well as in the name of the vector `mNetworkGameObjects`... at the end of the day once it's in that vector these are just state updates, do we really care they're from the network?

src/client/NetworkingClient.cpp
```diff
@@ -127,7 +127,7 @@ void NetworkingClient::RealtimeClientSend() {
         if (mAcceptingDirectionRead) {
             sf::Packet packet;
             sf::Uint32 move_cmd = (sf::Uint32)RealtimeCommand::move;
-            if (packet << mClientId << move_cmd << mCurrentTick << mDirection.x_dir << mDirection.y_dir) {
+            if (packet << mClientId << move_cmd << mCurrentTick << mDirection.x << mDirection.y) {
```
copacetic: this should probably have something like network_game_object right? serializedGameMove?

copacetic: Might be a rare case where inheritance is actually useful.

src/common/game_def.hpp
```diff
@@ -9,13 +9,15 @@ struct keys {
     sf::Keyboard::Key up, down, left, right;
 };
 
-struct position {
+struct network_game_object {
     sf::Uint32 id;
     float x_pos;
     float y_pos;
+    float size;
```
copacetic: things are getting mixed up. size is talking about the radius of a group, but the id is just a client id right now right?

copacetic: groups != clients is my thought.

copacetic: I wonder if we need to start designing for grouping/ungrouping. like maybe this should have a field added for "member clients" or member_players. I can imagine many kind of network_game_objects, such as mines or obstacles. Since we know these are part of the game we can probably save ourselves refactoring later, if we try to create something more general now. (usually Im against designing too generally too early, but since we know the game design, I think it will probably save us work overall)

src/server/GameController.cpp
```diff
@@ -22,33 +22,58 @@ GameController::~GameController() {
...
-    // Update players
-    mPlayers[client_id]->setDirection(client_direction);
+void GameController::update() {
+    // Get clients
+    std::vector<int> client_ids = mNetworkingServer->getClientIds();
+
+    // Create new players
+    for (const int client_id : client_ids) {
+        if (!mPlayers[client_id]->isActive()) {
+            createPlayer(client_id);
```
copacetic: why is createPlayer called when it's *not* active. surprising code

src/server/GameController.cpp
```diff
@@ -22,33 +22,58 @@ GameController::~GameController() {
...
+        }
+    }
+
+    // Deactivate players without a client
+    for (Player* player : mPlayers) {
```
copacetic: setting a player to inactive for even a second might have surprising results no? I wonder if this should be looking up in client_ids to decide whether to set player to inactive or active. It's inefficient at the moment because it's a vector, so no easy way to check membership efficiently. But the # of players is small so it doesn't really matter if we're doing O(n) lookups. eventually might want to consider an ordered hash.

src/server/GameController.cpp
```diff
@@ -22,33 +22,58 @@ GameController::~GameController() {
...
     mPlayers[new_player_id]->setActive(true);
-    mGroups[new_group_id]->setActive(true);
-
-    mGroups[new_group_id]->addMemeber(mPlayers[new_player_id]);
+    std::cout << "Created player " << new_player_id << std::endl;
 
-    return new_player_id;
+    // Create group for player
+    mGroups[new_player_id]->setActive(true);
+    mGroups[new_player_id]->addMemeber(mPlayers[new_player_id]);
```
copacetic: typo: Memeber -> Member

src/server/NetworkingServer.cpp
```diff
@@ -38,37 +79,19 @@ void NetworkingServer::RealtimeServer() {
...
-                    std::cout << "Sending game state to client: " << sender << " " << port << std::endl;
+                    //std::cout << "Sending game state to client: " << sender << " " << port << std::endl;
                     // sample current state every 100 ms, this simply packages and returns it
                     game_state_packet << ct; // should have error handling for <<
-                    for (const auto iter : mClientPositions) {
-                        sf::Uint32 client_id = iter.first;
-                        game_state_packet << client_id << iter.second[0] << iter.second[1];
+                    if (mAcceptingNetworkGameObjectsReads) {
+                        for(const auto network_game_object: mNetworkGameObjects) {
+                            game_state_packet << network_game_object.id
```
copacetic: shouldn't this be simply `game_state_packet << network_game_object`?

src/server/NetworkingServer.cpp
```diff
@@ -129,16 +152,8 @@ void NetworkingServer::ApiServer() {
...
-                                    sf::Packet response_packet;
-                                    sf::Uint32 api_command = (sf::Uint32)APICommand::register_client;
-                                    if(response_packet << api_command << mClientIdCounter << (sf::Uint32) mCurrTick) {
-                                        client.send(response_packet);
-                                        std::cout << "Received client registration. Issued client ID " << mClientIdCounter << std::endl;
-                                        mClientPositions[mClientIdCounter] = new float[2]{0.f, 0.f};
-                                        mClientSocketsToIds[&client] = mClientIdCounter;
-                                        mClientIdCounter++;
-                                    }
+                                if (packet >> api_command && api_command == (sf::Uint32)APICommand::register_client) {
```
copacetic: nice, thanks for pulling RegisterClient and Move into their own methods, much cleaner

src/server/NetworkingServer.cpp
```diff
@@ -150,6 +165,8 @@ void NetworkingServer::ApiServer() {
                                 std::cout << "TCP client disconnected. Removing client. " << std::endl;
                                 DeleteClient(&client, selector);
                                 break;
+                            default:
```
copacetic: were you getting a warning or something? this default is pretty pointless. at least we should put a `cout` saying this happened

src/server/NetworkingServer.cpp
```diff
@@ -163,36 +180,53 @@ void NetworkingServer::ComputeGameState() {
...
-                mClientPositions[client_id][1] += move_vector[1]*5.f; // y component
-            }
-            std::cout << iter.first << " " << iter.second[0] << " " << iter.second[1] << std::endl;
-        }
-        std::cout << "game state thread: dumping client positions for tick " << mCurrTick << std::endl;
-        for (const auto iter : mClientPositions) {
-            std::cout << iter.first << " " << iter.second[0] << " " << iter.second[1] << std::endl;
-        }
         // window is closed, time to reset. May want to use fix sized array as an optimization.
         // It's a bit mean to disguard old commands, maybe there should be windows for multiple ticks?
```
copacetic: wait all I see here is deleted code...  don't you need something to call something here? if this is simply the "game clock" then it should be renamed from computeGameState, to runGameClock or something

src/server/NetworkingServer.cpp
copacetic: aha yeah nice catch. previously when the computation was happening in this function, it was a given that it would be false at the right time. Im curious why you pulled the computation of game state out of this code

src/server/NetworkingServer.cpp
```diff
@@ -163,36 +180,53 @@ void NetworkingServer::ComputeGameState() {
...
+    if (mAcceptingMoveCommands == false) {
+        for(const auto& client_move : mClientMoves) {
+            network_player np = {client_move.first, client_move.second[0], client_move.second[1]};
+            network_players.push_back(np);
+        }
+    }
+    return network_players;
+}
+
+void NetworkingServer::setState(std::vector<Group*> active_groups) {
```
copacetic: eh I rather be as specific as possible. Maybe call this `setNetworkGameObjects`

src/server/main.cpp
copacetic: I dont get it, is the server just moving the clients by itself? in the initial design a single tick was equal to the time of accepting move commands (100ms) + time to compute game state. Now it seems artificial and detached from the client actions

copacetic:
Big change! Lots of positives here. I do feel like we should discuss architectural changes beforehand. not sure why the computeGameState thread is so different for example

src/server/NetworkingServer.cpp
SourenP: To achieve this, I think you need extend packets to support your custom types: https://www.sfml-dev.org/tutorials/2.5/network-packet.php

src/server/NetworkingServer.cpp
SourenP: Yeah I was getting a warning, yeah doing some error logging here would be good. Will add it.

src/server/NetworkingServer.cpp
SourenP: This is meant to be the single function that sets the entire state that needs to be retrieved by clients. Right now we only have groups that are converted to `network_game_objects` but in the future we will also be setting mines, resources, etc.

I don't mind keeping it `setState` so that it indicates how it's meant to be expanded.

src/client/NetworkingClient.cpp
SourenP: To achieve this, I think we need extend packets to support your custom types: https://www.sfml-dev.org/tutorials/2.5/network-packet.php
Didn't feel like doing that in this branch.

src/client/NetworkingClient.cpp
SourenP: Rather than passing just positions (id, x_dir, y_dir) over the network we are now passing data that describes a game object (essentially a circle). 
In the previous code we we had a hardcoded `size` value, now this is sent by the server - so if there is a group of size 2, it'll actually look big on the client.

src/common/game_def.hpp
SourenP: Right now, per client we have 1 group and 1 player, so I'm using client_id, group_id and player_id interchangeably. 
Of course this isn't going to work in the long term but I think this is a reasonable step forward for this change.

The code change that supports communicating these ids distinctly is non trivial.

src/server/NetworkingServer.cpp
SourenP: I don't want `NetworkServer` to call an update function in `GameController` from `ComputeGameState`. 
I don't want the `NetworkServer` to have a reference to `GameController` or even know of its existence.
I want the `NetworkServer` to be _used_ by the `GameController` to communicate state to and receive commands from the client.

Let's discuss this in https://github.com/SourenP/ungroup_sfml/pull/19/files/c5d2bc2e8609966792ecc862ac310ecdf5999c43#r289660647

src/server/main.cpp
SourenP: I'm confused...

So during the accepting move commands 100ms sleep, we should be computing game state?

src/server/GameController.cpp
copacetic: I was thinking of the case where a different thread might read this state

src/server/NetworkingServer.cpp
SourenP: Move the tick update to `GameController` in b5490ec

src/server/NetworkingServer.cpp
SourenP: Fixed

src/server/GameController.cpp
SourenP: Fixed in b5490ec

src/server/GameController.cpp
SourenP: That can't happen because we have the `mAcceptingNetworkGameObjectsReads` variable in `NetworkingServer::setState`

src/server/NetworkingServer.cpp
SourenP: Moved the game state update to Gamecontroller in b5490ec

src/server/main.cpp
SourenP: Moved the game state update to `Gamecontroller` in b5490ec

src/client/NetworkingClient.cpp
copacetic: I do think we can do better here wrt naming / abstraction, like why don't we have a GameObject which defines the << >> functions so it can be serialized and we don't have a distinction that way

copacetic: Im fine to commit since we're still early and it's not worth focusing on

src/client/NetworkingClient.cpp
copacetic: fair enough :) probably something to do down the line, but not a big deal

src/common/game_def.hpp
copacetic: fair enough, I'd put that in a comment `I'm using client_id, group_id and player_id interchangeably.`
and explain why, it's a little confusing

src/server/GameController.cpp
copacetic: it sort of makes sense, doesn't make me happy- can we think of a better way? Im ok shipping this if what you just wrote is in the code as a comment or in docs (which we don't have a system for lol). It's definitely hard to understand

src/server/GameController.cpp
copacetic: yeah... it's an implicit dependency. code changes it's weird that a variable called `mAcceptingNetworkGameObjectReads` is gating something about players. in fact the NGO reads are coming over the UDP socket, the player data is more about the TCP registration/connection (that's where players are removed when the TCP connection dies)

src/server/NetworkingServer.cpp
copacetic: I'd put that in a comment. I worry that we make assumptions about the future that often don't end up true... because it's the unknowable future. I respect that at some point that's how this might end up, but we're not there and I'd prefer clarity now over one less search and replace later if that makes sense.

PR #20, opened 2019-06-06
Abstract collision handling and detection

SourenP:
- Make GroupShape base class of Group
- Add velocity to GroupShape
- Abstract CollisionDetector to handle GroupShape instead of Group
- Create templated collision handling function for Group


src/common/GroupShape.cpp
```diff
@@ -27,11 +33,22 @@ float GroupShape::getRadius() const {
...
 
-
 void GroupShape::setRadius(int size) {
     mCircle.setRadius(size);
 }
+
+void GroupShape::move() {
+    mCircle.setPosition(mCircle.getPosition() + mVelocity);
+}
+
```
copacetic: I'd add a comment explaining why we need this method

src/server/CollisionDetector.hpp
```diff
@@ -3,12 +3,28 @@
...
     public:
         CollisionDetector();
-        void detectGroupCollisions(std::vector<Group*> groups);
+        template<typename T> void detectAndHandleShapeCollisions(const std::vector<T*>& shapes);
+        void handleGroupShapeCollision(GroupShape* group_a, GroupShape* group_b);
+        void detectAndHandleGroupShapeCollisoin(std::vector<GroupShape*> group_shapes);
 };
 
+// TODO: Move this into a tpp file?
+template<class Group> void CollisionDetector::detectAndHandleShapeCollisions(const std::vector<Group*>& shapes) {
```
copacetic: this line is confusing af

copacetic: the casting feels unnecessary. seems like renderable state like color should be wrapped around the physics object. like I think the client can have a RenderableGroup with a reference to a PhysicsGroup, the collision detector only care about PhysicsGroups, so we only ever pass it those. no type casting or weird templates

src/server/CollisionDetector.cpp
```diff
@@ -1,24 +1,35 @@
...
 
 CollisionDetector::CollisionDetector() {
     //
 }
 
-// TODO: Move collision handling out and make groups param const
-void CollisionDetector::detectGroupCollisions(std::vector<Group*> groups) {
-    for(Group* group_a: groups) {
-        for(Group* group_b: groups) {
+void CollisionDetector::handleGroupShapeCollision(GroupShape* group_a, GroupShape* group_b) {
```
copacetic: this logic could use a comment or documentation

src/common/collision.cpp
```diff
@@ -0,0 +1,31 @@
+#include "collision.hpp"
+#include <iostream>
+
+void handle_circle_collision(std::vector<Circle*>& circles) {
```
copacetic: We might want this to be a static function on the Group class that actually looks like `Group.handleCollisions(std::vector<Group*>& groups)`, then you pass it a vector of groups and it knows that groups are represented by circles so it does the collision using the circle object of the group. that way you dont have to make a new vector for every single group every time you call this function

src/server/GameController.cpp
```diff
@@ -67,7 +66,11 @@ void GameController::updateState() {
     }
 
     // Detect and handle group collisions
-    mCollisionDetector->detectAndHandleShapeCollisions(mGroups);
```
copacetic: this part got a little uglier / more inefficient, that's why I suggested the alternative way in my other comment

src/client/GroupCircle.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef GroupCircle_hpp
+#define GroupCircle_hpp
+
+#include <SFML/Graphics.hpp>
+#include <stdio.h>
+#include "../common/Circle.hpp"
+#include "../common/GameObject.hpp"
+
+class GroupCircle : public GameObject {
```
copacetic: Niiiiice, I way prefer this way. One thing Im a little confused about is the naming here. Should this just be called Group?  I'm actually not sure why it exists also. Is it just to hold the group ID? if it's going to be the Group class, Im guessing down the line it can hold a list of member clients?

SourenP: This is the representation of a Group for the client - it's just a circle representing a group. 
I think we should leave the design this way for now since it's unclear how much the client will actually need to know in the future.
With this we can do velocity and collision interpolation on the client.

If you can think of a better name, I'm down to change it.

SourenP: The actual `Group` class is in the server (`Group.hpp`)

src/common/collision.cpp
SourenP: Sure I'm down.

SourenP: Sure I'm down

src/common/GroupShape.cpp
SourenP: ? seems pretty straightforward to me

src/common/collision.cpp
SourenP: Done in 9d5a236

src/server/CollisionDetector.cpp
SourenP: Done in 9d5a236

src/common/GroupShape.cpp
copacetic: Why is there the one that takes an offset was my q

src/client/GroupCircle.hpp
copacetic: I know Im annoying, but my fundamental question is: why aren't the group representations pretty much the same on server and client? Why aren't both of these called Group?

SourenP: Because the client doesn't need all the functionality of `Group`, at least not for now. We want to avoid sending all the data and writing all the code needed to update/reconstruct a group.

The representation of a Group on the client will be clearer as we go along, but for the foreseeable future all we need is circle that we can interpolate movement and collision for.

imo

src/common/GroupShape.cpp
SourenP: Oh, the offset one is used by the collision handler, this one is used by the group's update method.

copacetic: that's the comment I was thinking

src/client/GroupCircle.hpp
copacetic: Aha, then maybe we shouldn't be sending groups, we should be sending circles?

SourenP: Well the client needs to know that it's receiving a Group not just a circle.
What data we send exactly is a discussion for https://github.com/SourenP/ungroup_sfml/pull/19

Would like to figure out how we want clients to interpret/represent Groups.

copacetic: sounds good, so we basically are just calling this GroupCircle to avoid a naming collision with the name Group on server side, but they're basically the same object?

PR #24, opened 2019-06-21
Code cleanup

SourenP:


src/server/NetworkingServer.hpp
```diff
@@ -21,36 +21,41 @@ class NetworkingServer {
...
-        // Methods
         void RealtimeServer();
         void ApiServer();
-
         void DeleteClient(sf::TcpSocket* client, sf::SocketSelector selector);
         void Move(sf::Packet command_packet, sf::Uint32 client_id, sf::Uint32 tick);
         void RegisterClient(sf::TcpSocket& client);
 
-        // Variables
+        // Thread safe mClientGroupUpdates operations
```
SourenP: Would be nicer if this was implemented in a thread safe vector class for mClientGroupUpdates
Ditto for mClientMove and mClientSocketsToIds (unordered_maps)

src/client/Client.cpp
```diff
@@ -33,15 +32,16 @@ void Client::draw(sf::RenderTarget& target) {
...
     // Get group circle updates
-    std::vector<group_circle_update> group_circle_updates = mNetworkingClient->getClientGroupUpdates();
+    std::vector<client_group_update> client_group_updates = mNetworkingClient->getClientGroupUpdates();
     std::vector<int> client_ids;
-    for (const auto group_circle_update : group_circle_updates) {
-        client_ids.push_back(group_circle_update.client_id);
-    }
+    std::transform(
+        client_group_updates.begin(), client_group_updates.end(), std::back_inserter(client_ids),
+        [](client_group_update cgu){return cgu.client_id;}
```
copacetic: what is this a lambda?

src/client/ClientGroup.hpp
```diff
@@ -9,8 +9,10 @@
 class ClientGroup : public GameObject {
 
     public:
-        ClientGroup(sf::Vector2f position);
+        explicit ClientGroup(sf::Vector2f position);
```
copacetic: why was this change required?

src/client/NetworkingClient.cpp
```diff
@@ -9,17 +9,9 @@
...
-    mAcceptingDirectionRead = true;
-    mDirection = sf::Vector2f(0.f, 0.f);
-}
-
-NetworkingClient::~NetworkingClient() {
-}
-
-sf::Uint32 NetworkingClient::Start() {
+NetworkingClient::NetworkingClient()
+    :mAcceptingNetworkGameObjectsRead{true}, mAcceptingDirectionRead(true), mDirection(0.f, 0.f)
```
copacetic: what's with the {true}

src/client/Client.cpp
SourenP: Given a `client_group_update` struct it returns its `client_id` field.

src/client/NetworkingClient.cpp
```diff
@@ -42,14 +34,16 @@ sf::Uint32 NetworkingClient::Start() {
     RealtimeClientRecv_thread.detach();
     RealtimeClientSend_thread.detach();
     SyncServerState_thread.detach();
-    return mClientId;
 }
 
-std::vector<group_circle_update> NetworkingClient::getClientGroupUpdates() {
+NetworkingClient::~NetworkingClient() {
```
copacetic: might as well put it on oneline

src/server/GameController.cpp
```diff
@@ -29,16 +28,12 @@ void GameController::update() {
...
-
-    // Read client inputs
-    return mNetworkingServer->getClientInputs();
+    return mNetworkingServer->collectClientInputs();
 }
 
 void GameController::computeGameState(
-    std::vector<int> client_ids,
-    std::vector<client_direction_update> client_direction_updates
+    std::vector<int>& client_ids,
```
copacetic: why change these to be references

src/server/Group.cpp
```diff
@@ -16,32 +18,40 @@ Group::~Group() {
...
-            setActive(true);
-            return;
-        }
-    }
-    setActive(false);
+    bool any_active_members = std::any_of(
+        mMembers.begin(), mMembers.end(),
+        [](std::shared_ptr<Player> player){return player->isActive();}
+    );
+    if (any_active_members)
```
copacetic: add the curly braces here, not having them is bad style

src/server/NetworkingServer.cpp
```diff
@@ -1,71 +1,107 @@
 #include "NetworkingServer.hpp"
 
-NetworkingServer::NetworkingServer() {
-    mClientIdCounter = 0;
-    mCurrTick = 0;
+NetworkingServer::NetworkingServer():mCurrTick(0) {
```
copacetic: I don't know why mCurrTick was moved here, this is a lot harder to read

src/server/NetworkingServer.cpp
```diff
@@ -109,6 +143,27 @@ void NetworkingServer::RealtimeServer() {
     }
 }
 
+
+void NetworkingServer::Move(sf::Packet command_packet, sf::Uint32 client_id, sf::Uint32 tick) {
+
```
copacetic: extra newline

src/server/NetworkingServer.hpp
copacetic: maybe we should just use a third party library for that

src/client/NetworkingClient.cpp
SourenP: read this: https://stackoverflow.com/questions/27352021/c11-member-initializer-list-vs-in-class-initializer/27354688

src/server/GameController.cpp
SourenP: Code-climate (cppcheck) pointed out: "Parameter client_ids is passed by value. It could be passed as a const reference which is usually faster and recommended in C++."

Basically doesn't need to copy the whole vector (which is faster).

src/server/Group.cpp
SourenP: ok

src/server/NetworkingServer.cpp
SourenP: Codeclimate (cppcheck) complained about this being constructed in the NetworkingServer constructor rather than before. We can't initialize it in the header file AFAICT because "You cannot use copy-initialization, because std::atomic_bool is not copy-constructible" so I have to do an inline list _direct_ initialization.

See https://stackoverflow.com/a/15751048

Read this: https://stackoverflow.com/a/15751048

src/client/ClientGroup.hpp
SourenP: Codeclimate (cppcheck) complained about this.

Here is a good explanation about why this is good: https://stackoverflow.com/a/121163

src/server/NetworkingServer.hpp
SourenP: Yeah maybe, I"ll look around. 
Going to leave it for another PR.

PR #44, opened 2019-08-03
Physics controller

SourenP:
- Define CircleRigidBody
CircleGameObjects now own a Circle (for rendering, ought to rename) and a CircleRigidBody (for physics)

- Define PhysicsController
The physics controller is in charge of stepping (moving) rigid bodies and handling collision between them.
 It is owned by the GameController.
It owns a vector of pointers to all rigid bodies in the game.

- Update GameController
The game controller now does this each step:
1. Updates game objects
2. Takes a physics controller step
3. Handles collision 
4. Updates game objects to match their rigid bodies (this step is a bit clunky, looking to improve)

I recommend chapter “13.5.2 Updating the Simulation” from the “Game Engine Architecture” book.
6337351C-A8BF-4138-B714-617C9E6205CE


src/common/Circle.hpp
```diff
@@ -12,17 +12,17 @@ class Circle {
...
-        sf::Vector2f getVelocity() const;
-
-        // Setters
         void setPosition(sf::Vector2f position);
+
+        float getRadius() const;
         void setRadius(int size);
+
+        // TODO(souren): On client use rigid body instead of these to move circles
+        void move();
```
SourenP: Delete these undefined methods

src/server/GameController.cpp
```diff
@@ -93,32 +115,21 @@ void GameController::updatePlayers(std::vector<client_direction_update> client_d
...
-    // Concat active group and mine circles
-    active_circles.reserve(active_group_circles.size() + active_mine_circles.size());
-    active_circles.insert(
-        active_circles.end(), active_group_circles.begin(), active_group_circles.end());
-    active_circles.insert(
-            active_circles.end(), active_mine_circles.begin(), active_mine_circles.end());
-
-    // Detect and handle collision
-    handle_circle_collision(active_circles);
+void GameController::matchRigid() {
```
SourenP: This step could be done in a cleaner way maybe. Let me know if you have any thoughts.
It can at least have a better name ;)

src/server/GameController.cpp
```diff
@@ -49,9 +61,19 @@ void GameController::computeGameState(
     const std::vector<int>& client_ids,
     const std::vector<client_direction_update>& client_direction_updates
 ) {
+    // Update game objects
     refreshPlayers(client_ids);
     updatePlayers(client_direction_updates);
-    refreshAndUpdateGroups();
+    updateGroups();
```
SourenP: It would be nice if we iterated over all GameObjects and called update rather than doing it individually for Groups, Mines, etc

src/common/Circle.hpp
SourenP: Fixed in 9ad68ad

src/server/GameController.cpp
SourenP: Not sure how to do this yet... we would have to store a vector of GameObjects instead of Groups and Mines.
Will leave for another PR.

src/server/GameController.cpp
SourenP: Fixed in b69748c

src/common/CircleGameObject.cpp
```diff
@@ -1,18 +1,38 @@
 #include "CircleGameObject.hpp"
 
 
-CircleGameObject::CircleGameObject(unsigned int id, sf::Vector2f position, float size)
-    :GameObject(id), mCircle(std::shared_ptr<Circle>(new Circle(size, position))) {}
+CircleGameObject::CircleGameObject(
+    unsigned int id, sf::Vector2f position, float radius, std::shared_ptr<CircleRigidBody> crb):
```
SourenP: Maybe rather than passing in a CircleRigidBody, we should pass a reference of the PhysicsContoller and have this class initialize it's own CircleRigidBody via createCRB()

src/client/Client.cpp
```diff
@@ -4,23 +4,25 @@
...
     }
 
     // Set up input
     mKeys.up = keys[0];
     mKeys.down = keys[1];
     mKeys.right = keys[2];
     mKeys.left = keys[3];
-
-    // Networking
-    mNetworkingClient = new NetworkingClient();
```
copacetic: where does this move to?

src/common/CircleGameObject.cpp
SourenP: Made this change 93b3e79

src/common/CircleGameObject.cpp
```diff
@@ -1,18 +1,35 @@
...
 
-std::vector<std::shared_ptr<Circle>> CircleGameObject::getCircles(
-    std::vector<std::shared_ptr<CircleGameObject>>& circle_game_objects) {
-    std::vector<std::shared_ptr<Circle>> circles;
-    std::transform(
-        circle_game_objects.begin(), circle_game_objects.end(), std::back_inserter(circles),
-        [](std::shared_ptr<CircleGameObject> cgo){return cgo->getCircle();});
-    return circles;
-};
+void CircleGameObject::setActive(bool is_active) {
```
copacetic: mind adding a comment regarding what active means here?

src/common/CircleGameObject.cpp
```diff
@@ -1,18 +1,35 @@
...
+    mIsActive = is_active;
+    mCircleRigidBody->setActive(is_active);
+}
+
+void CircleGameObject::setVelocity(sf::Vector2f velocity) {
+    mCircleRigidBody->setVelocity(velocity);
```
copacetic: so is the circle rigid body shared by both the physics controller and the circle game object? Seems like we might end up leaking memory if we're not careful

copacetic: I guess I'm wondering if the physics controller should be the only manager and when we need to update a certain circles properties, we call the physics controller with that circles ID. single owner seems less error prone and easier to reason about

src/common/CircleGameObject.cpp
```diff
@@ -1,18 +1,35 @@
...
+    mCircle->setRadius(radius);
+    mCircleRigidBody->setRadius(radius);
+}
+
+void CircleGameObject::setPosition(sf::Vector2f position) {
+    mCircle->setPosition(position);
+    mCircleRigidBody->setPosition(position);
+}
+
+void CircleGameObject::matchRigid() {
```
copacetic: when is this necessary? worth commenting since it's surprising at first glance. I'm assuming maybe in interpolation?

src/common/CircleGameObject.hpp
```diff
@@ -3,18 +3,26 @@
...
+     CircleGameObject(unsigned int id, sf::Vector2f position, float radius, sf::Color color,
+        std::shared_ptr<PhysicsController> pc);
+     std::shared_ptr<Circle> getCircle();
+     void setActive(bool is_active);
+     void setVelocity(sf::Vector2f velocity);
+     void setRadius(float radius);
+     void setPosition(sf::Vector2f position);
+     void matchRigid();
+ protected:
+     std::shared_ptr<Circle> mCircle;
```
copacetic: what is this actually shared with?

copacetic: I mean (mCircle)

src/common/PhysicsController.cpp
```diff
@@ -0,0 +1,55 @@
...
+std::shared_ptr<CircleRigidBody> PhysicsController::createCRB(float radius, sf::Vector2f position) {
+    std::shared_ptr<CircleRigidBody> crb = std::shared_ptr<CircleRigidBody>(
+        new CircleRigidBody(radius, position));
+    mCircleRigidBodies.push_back(crb);
+    return crb;
+}
+
+void PhysicsController::step() {
+    for (auto crb : mCircleRigidBodies) {
+        crb->step();
```
copacetic: is it important that the steps happen "at once" I  can imagine half the circles being updated right when the frame updates or something. probably not a real issue, just thinking out loud. safe to ignore

src/common/PhysicsController.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+    Handles collision between CircleRigidBodies by pairs.
+    To handle the collision between circle A and B, circle B is moved in the direction
+    of the vector spanning from the center of cirlce A to the center of circle B by the
+    amount needed such that the circles are touching but not intersecting.
+*/
+void PhysicsController::handleCollision() {
+    std::vector<std::shared_ptr<CircleRigidBody>> active_circles;
+    std::copy_if(
```
copacetic: im curious if this act of copying is actually more efficient. copying is super expensive

copacetic: if the goal is to copy these as an optimization, i think it may be premature and not actually beneficial

src/common/PhysicsController.cpp
```diff
@@ -0,0 +1,55 @@
...
+    for (std::shared_ptr<CircleRigidBody> circle_a : active_circles) {
+        for (std::shared_ptr<CircleRigidBody> circle_b : active_circles) {
+            if (circle_a == circle_b) {
+                continue;
+            }
+            // Detect collision
+            float dist = distance(circle_a->getPosition(), circle_b->getPosition());
+            if (dist < circle_a->getRadius() + circle_b->getRadius()) {
+                // Handle collision
+                sf::Vector2f between = getVector(circle_a->getPosition(), circle_b->getPosition());
```
copacetic: what is getVector?

copacetic: vague name

src/common/PhysicsController.cpp
```diff
@@ -0,0 +1,55 @@
...
+                float radius_sum = circle_a->getRadius() + circle_b->getRadius();
+                float distance_between = length(between);
```
copacetic: what is length()? I don't understand `length(between)`. I think something here needs renaming

src/server/GameController.cpp
```diff
@@ -101,39 +123,28 @@ void GameController::updatePlayers(std::vector<client_direction_update> client_d
...
-    // Concat active group and mine circles
-    active_circles.reserve(active_group_circles.size() + active_mine_circles.size());
-    active_circles.insert(
-        active_circles.end(), active_group_circles.begin(), active_group_circles.end());
-    active_circles.insert(
-            active_circles.end(), active_mine_circles.begin(), active_mine_circles.end());
-
-    // Detect and handle collision
-    handle_circle_collision(active_circles);
+void GameController::updateGameObjectsPostPhysics() {
```
copacetic: nice

src/tests/common/tests-common.cpp
```diff
@@ -1,19 +1,19 @@
...
+      circle->step();
       REQUIRE(circle->getPosition() == sf::Vector2f(10.f, 10.f));
     }
-    WHEN( "move is called many times" ) {
+    WHEN( "step is called many times" ) {
       for (int i=0; i < 10; i++) {
-        circle->move();
+        circle->step();
       }
       REQUIRE(circle->getPosition() == sf::Vector2f(10*10.f, 10*10.f));
```
copacetic: can you update these tests to check the rigid bodies locations? I think it's time we require tests in our CRs :(

src/common/CircleGameObject.cpp
SourenP: physics controller is the owner, that's where it's initialized with `new` and kept in a vector `mCircleRigidBodies`. 
CircleGameObject just has a pointer to that object. Ideally it wouldn't even be allowed to delete it, but it's not really its job to delete it.
Just in case I've made the pointer a `std::shared_ptr<CircleRigidBody>` which is a "smart" pointer, but to be honest I haven't looked into this enough to be certain. 
At some point I really need to look into valgrind and define reconstructors.

That being said I don't think this design is flawed.

SourenP: It would be verbose and non-sustainable to create methods in PhysicsController with which you modify `CircleRigidBody`s... that's why I have the pointer shared so `CircleGameObject` can call things like `setVelocity`.

src/common/PhysicsController.cpp
SourenP: We set the network state _after_ we've updated the game state. The physics controller step is a substep of "update game state". 
What you're thinking about isn't a concern.

src/common/PhysicsController.cpp
SourenP: You mean you'd prefer to just check if `circle_a` and `circle_b` are active in the loop below? 
Is that really faster/cleaner? Note that we're copying pointers not entire objects.

src/common/PhysicsController.cpp
SourenP: I'm going to redo collision in another review, there is an bug I want to fix: https://github.com/ungroup/ungroup_game/issues/43
I'll take this comments into account there

src/common/PhysicsController.cpp
SourenP: I'm going to redo collision in another review, there is an bug I want to fix: https://github.com/ungroup/ungroup_game/issues/43
I'll take this comments into account there

src/common/CircleGameObject.hpp
SourenP: I made this "shared" because there is this function in `CircleGameObject`
```
std::shared_ptr<Circle> CircleGameObject::getCircle() {
    return mCircle;
}
```

src/common/CircleGameObject.cpp
SourenP: So after the physics world updates the rigid bodies (applies velocity and corrects collision) the `CircleGameObject`s visual representation of the circle, needs to be updated to match the physical representation of the circle.

Substeps per step of GameController
> Updates game objects
> Takes a physics controller step
> Handles collision
> Updates game objects to match their rigid bodies (this step is a bit clunky, looking to improve)

src/common/CircleGameObject.cpp
SourenP: Will add a comment

src/tests/common/tests-common.cpp
SourenP: Not ready to write tests yet :(

src/common/CircleGameObject.cpp
copacetic: sounds good, I think it's fine for now, if we see errors we can reconsider

src/common/PhysicsController.cpp
copacetic: yes  just checking active state. allocating memory and copying stuff is a lot more expensive than checking a boolean (few asm instructions, vs a call to the kernel to get more memory possibly)

copacetic: anyway since this is the way it is, I think it's fine to leave for now, we'll want to rewrite collision detection with a better data structure eventually anyway so that it's not n^2

src/tests/common/tests-common.cpp
copacetic: what do you mean

PR #45, opened 2019-08-11
Add groupability

copacetic:


src/makefile
```diff
@@ -9,7 +9,7 @@ endif
 
 ifeq ($(detected_OS),Darwin)        # Mac OS X
 	CXX = clang++
-	CXXFLAGS = -std=c++11 -stdlib=libc++
+	CXXFLAGS = -std=c++11 -stdlib=libc++ -g
```
copacetic: enables debug information so that gdb can have line numbers etc.
we should add production/devel/staging build options eventually and this should be excluded from prod builds blah blah blah

src/server/Group.cpp
```diff
@@ -25,6 +25,17 @@ void Group::update() {
...
         mCircle->setVelocity(new_velocity * GROUP_SPEED);
 
+        // group is groupable if any member player wants to group
+        // should probably switch to voting functionality later
+        bool groupable = std::accumulate(
+            mMembers.begin(),
+            mMembers.end(),
+            false,
+            [](bool curr, std::shared_ptr<Player> player) {
+                return curr || + player->getGroupable();
```
copacetic: woah typo the + shouldn't be there. wtf this works o_0

src/server/NetworkingServer.cpp
```diff
@@ -142,9 +142,14 @@ void NetworkingServer::HandleApiCommand(
...
         case sf::Socket::Done:
-            if (command_packet >> api_command_type &&
-                api_command_type == (sf::Uint32)APICommandType::register_client) {
+            if (command_packet >> api_command_type) {
+              if (api_command_type == (sf::Uint32)APICommandType::register_client) {
                 RegisterClient(client);
+              }
+
+              if (api_command_type == (sf::Uint32)APICommandType::group) {
+                SetGroupable(client);
```
copacetic: should be indented more. not sure why it's capitalized ...

src/server/NetworkingServer.cpp
```diff
@@ -165,10 +170,30 @@ void NetworkingServer::HandleApiCommand(
 
 void NetworkingServer::DeleteClient(sf::TcpSocket* client, std::list<sf::TcpSocket*> clients) {
```
copacetic: @SourenP we gotta get more consistent. in some place functions_are_snake_case in others they are camelCase and here they're SomethingElse

copacetic:
kinda blew my mind how many different files I had to touch to add one boolean... 

Im not sure how to make this better though. Maybe a shared file between client and server which describes the networking state? protobuf does this I think, we could try using that.

OTOH a lot of files which were touched were random classes which contain classes which contain classes and every one of them needed a setter/getter. I almost wonder if something like GraphQL would help with this? I think it's just this architecture

src/client/Client.cpp
```diff
@@ -163,6 +173,10 @@ sf::Uint32 Client::getId() const {
 
 void Client::handleEvents(sf::Event& event) {
     if (event.type == sf::Event::KeyPressed) {
+        if (sf::Keyboard::isKeyPressed(mKeys.group)) {
+            mGroupable ^= true;
```
SourenP: why do we need an XOR here?

src/client/Client.cpp
```diff
@@ -29,7 +30,14 @@ void Client::draw(sf::RenderTarget& target, sf::Shader* shader, bool use_shader)
     // Draw active client groups
     for (auto client_group : mClientGroups) {
         if (client_group->isActive()) {
-            client_group->getCircle()->draw(target, shader, use_shader);
+            bool groupable = client_group->getGroupable();
+            std::shared_ptr<Circle> circle = client_group->getCircle();
+            if (groupable) {
+                circle->setColor(sf::Color(255, 0, 0));
+            } else {
+                circle->setColor(sf::Color(0, 0, 255));
```
SourenP: It makes sense for us to change the color or apply a shader on the client based on `groupable` but the default color of circles should be stored/set internally on the *server* not the client. With your change here, it's impossible to have different color circles for each client.

src/client/NetworkingClient.cpp
```diff
@@ -21,11 +21,9 @@ NetworkingClient::NetworkingClient():mDirection(0.f, 0.f) {
 
     std::cout << "Starting ungroup demo client." << std::endl;
 
-    /*
     // api
-    std::thread api_client_recv_thread(api_client_recv, &api_client);
-    std::thread api_client_send_thread(api_client_send, &api_client);
-    */
+    std::thread ApiClientRecv_thread(&NetworkingClient::ApiClientRecv, this);
+    std::thread ApiClientSend_thread(&NetworkingClient::ApiClientSend, this);
```
SourenP: Yay we're using the API Cleint :D

src/client/NetworkingClient.cpp
```diff
@@ -86,6 +91,34 @@ void NetworkingClient::RegisterNetworkingClient() {
...
 }
 
+void NetworkingClient::ApiClientRecv() {
+    while (true) {
+        sf::Packet group_response;
+        ApiCommand api_command;
+
+        if (mApiClient->receive(group_response) == sf::Socket::Done) {
+            if (group_response >> api_command &&
+                api_command.command == (sf::Uint32)APICommandType::group) {
```
SourenP: Calling this Command `group` is confusing and vague

src/client/NetworkingClient.cpp
```diff
@@ -53,6 +53,11 @@ void NetworkingClient::setDirection(sf::Vector2f direction) {
     mDirection.set(direction);
 }
 
+void NetworkingClient::setGroupable(bool groupable) {
+    mNeedsGroupStateSync = mGroupable != groupable;
```
SourenP: what's this variable for?

src/common/Circle.cpp
```diff
@@ -7,9 +7,8 @@
 
 Circle::Circle(float size, sf::Vector2f position):mCircleShape(size) {
     mCircleShape.setPosition(position);
-    unsigned int seed = time(NULL);
-    mCircleShape.setFillColor(
-        sf::Color(rand_r(&seed) % 255, rand_r(&seed) % 255, rand_r(&seed) % 255));
+    mColor = sf::Color(0, 0, 255);
```
SourenP: mCircleShape has a color property, why dont you just uses that to store the color
https://www.sfml-dev.org/tutorials/2.0/graphics-shape.php

src/server/NetworkingServer.hpp
```diff
@@ -57,6 +59,7 @@ class NetworkingServer {
      ThreadSafeVector<GroupUpdate> mGroupUpdates;
      ThreadSafeVector<MineUpdate> mMineUpdates;
      ThreadSafeMap<sf::Uint32, sf::Vector2f> mClientMoves;
+     ThreadSafeMap<sf::Uint32, bool> mClientGroupable;
```
SourenP: What do you think about combining mClientGroupable and mClientMoves into one map that stores all the updates for a client

src/client/Client.cpp
copacetic: xor just toggles the current value of the boolean.

1 ^ 1 = 0
0 ^ 1 = 1

src/client/Client.cpp
copacetic: previously it was just assigning a random color at boot and it wasn't communicating the color to the server

i agree with you the server should keep the colors or do something clever using the client_id to determine the color.... (client_id indexed into a pre-shared color array?)

but it's also a separate change, i dont think it's in-scope for this PR

src/client/NetworkingClient.cpp
copacetic: woot!

src/client/NetworkingClient.cpp
copacetic: alternative? maybe it should be toggle_group?

src/client/NetworkingClient.cpp
copacetic: the api client needs to know when to make an API request to the server, so that it doesn't spam the server. It basically should only make a request when there's a state transition (mGroupable != groupable), that is what's captured here

src/common/Circle.cpp
copacetic: good point, looks like I can call `getFillColor` to grab it

src/server/NetworkingServer.hpp
copacetic: not sure honestly. seems like it might be more complicated than simplifyng at this point

im thinking we need a better overall solution for synced state from the server

Im seeing a lot of similarities with web UI's and solutions like GraphQL. I dont feel ready yet to make changes till we're pretty convinced of the right way

copacetic: does your game dev book talk about this?

src/server/Group.cpp
SourenP: fix plox

src/server/NetworkingServer.cpp
SourenP: fix plox

src/server/NetworkingServer.cpp
SourenP: Please fix anything you notice

I'm naming member functions and variables camelCase and inline variables, arguments and parameters in snake_case

src/client/Client.cpp
SourenP: nice

src/client/NetworkingClient.cpp
SourenP: `toggle_groupability` sgtm

src/client/NetworkingClient.cpp
SourenP: Make sense. Please add a comment explaining this

src/common/Circle.cpp
SourenP: Yeah let's use that so that we use less variables/code

src/server/NetworkingServer.hpp
SourenP: ok lets leave it separate for now

SourenP:
> kinda blew my mind how many different files I had to touch to add one boolean...
> 
> Im not sure how to make this better though. Maybe a shared file between client and server which describes the networking state? protobuf does this I think, we could try using that.
> 
> OTOH a lot of files which were touched were random classes which contain classes which contain classes and every one of them needed a setter/getter. I almost wonder if something like GraphQL would help with this? I think it's just this architecture

Let's try to figure this out outside of this CR. Now that you've tried to add a network-shared variable you'll have more opinions about the (need for) solution.

SourenP:
It's kinda weird that this CR doesn't actually add any functionality. Sort of a half baked change.
Maybe *before* merging this to master, you can make another branch off of this one that completes the groupability feature.

src/common/Circle.cpp
SourenP: fixed in 92f3e38

src/client/NetworkingClient.cpp
SourenP: fixed in 92f3e38

src/client/Client.cpp
SourenP: fixed in 92f3e38

SourenP: Color is not being synced from the server to the client

PR #52, opened 2019-08-22
Fix smoke test

copacetic:
500 milliseconds is a flaky value. Timeout should be 1 second
to void flip floppitiness.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,12 @@
-env:
-  global:
-   # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created
-   #   via the "travis encrypt" command using the project repo's public key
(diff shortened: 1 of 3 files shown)


SourenP:
Why did you make a test in node?
Can't this be done in cpp using catch?

Also isn't it flaky to check the output of the executable?
What if we change the string?

copacetic:
I mentioned this in WhatsApp, but node is actually amazing for this because
it has asynchronous coding paradigms baked in, easy callbacks and streams
etc. It's also very fast to write JS instead of c++.

I also talked about the flakiness in WhatsApp 😂

SourenP:
ship it?

copacetic:
Hehe, we can revert it if you want, I developed on master by accident. You
should try using it

PR #65, opened 2019-09-08
Circle testing

vahaknp:
This PR does a few things:
1. Updates the `makefile` to catch all test files in `/tests/common/` instead of specifying each file one by one, or having all tests in one file
2. Renames the `common` test file to be circle specific
3. Change testing style to be more BDD (we can change this if we don't like it) and add more tests
4. Assert that the `x` and the `y` of the vector are correct, so that test failures are clear. See example:

Before:
```
tests/common/tests-circle.cpp:80: FAILED:
  REQUIRE( position == sf::Vector2f(10.f, 10.f) )
with expansion:
  {?} == {?}
```

After:
```
tests/common/tests-circle.cpp:80: FAILED:
  REQUIRE( position.x == number_of_steps*10.f )
with expansion:
  0.0f == 20.0f
```


SourenP:
lgtm 👍 
first push 🎉

PR #66, opened 2019-09-09
Event system

SourenP:
- Create an event system that uses a queue and callbacks
- Use event system to simplify code by creating/removing client players on connect/disconnect

- Update makefile to include all subdirectories of `SRC_DIR`. This was needed to link `src/common/events/*.cpp`


SourenP:
I need to make the EventController thread safe.

SourenP:
Used this singleton pattern https://stackoverflow.com/questions/1008019/c-singleton-design-pattern

src/common/events/EventController.cpp
```diff
@@ -0,0 +1,43 @@
...
+    mEventMapLock.unlock();
+}
+
+void EventController::queueEvent(std::shared_ptr<Event> event) {
+    mEventQueueLock.lock();
+    mEventQueue.push(event);
+    mEventQueueLock.unlock();
+}
+
+void EventController::forceProcessEvents() {
```
copacetic: shouldn't this function take locks since it is modifying mEventQueue?

src/makefile
```diff
@@ -20,17 +20,21 @@ CMN_DIR = ./common
 
 CLIENT_LDFLAGS = -lsfml-graphics -lsfml-window -lsfml-system -lsfml-audio -lsfml-network -lpthread
 CLIENT_SRC_DIR = ./client
-CLIENT_SRC_LIST = $(wildcard $(CMN_DIR)/*.cpp) $(wildcard $(CLIENT_SRC_DIR)/*.cpp)
+CLIENT_SRC_LIST = $(shell find $(CLIENT_SRC_DIR) $(CMN_DIR) -name "*.cpp")
```
copacetic: nice I like this shell find stuff

src/server/NetworkingServer.cpp
```diff
@@ -168,10 +175,11 @@ void NetworkingServer::handleReliableCommand(sf::Socket::Status status, sf::Pack
 }
 
 void NetworkingServer::deleteClient(sf::TcpSocket* client, std::list<sf::TcpSocket*> clients) {
-    int client_id = mClientSocketsToIds.get(client);
+    int deleted_client_id = mClientSocketsToIds.get(client);
     clients.remove(client);
     mClientSocketsToIds.erase(client);
-    mRemovedClientIds.add(client_id);
+    EventController::getInstance().queueEvent(
```
copacetic: I dont really understand shared_ptr's but we use em everywhere

src/tests/common/events/tests-EventController.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+void add_to_the_global(std::shared_ptr<Event> event) {
+    std::shared_ptr<TestEvent> test_event = std::dynamic_pointer_cast<TestEvent>(event);
+    THE_GLOBAL += test_event->getNumber();
+}
+
+SCENARIO("EventController processes an event", "[events]") {
+    GIVEN("an event listener is added") {
+        std::function<void(std::shared_ptr<Event>)> add_to_the_global_ptr = add_to_the_global;
+        EventController::getInstance().addEventListener(EventType::EVENT_TYPE_TEST,
```
copacetic: this should be indented?

src/tests/common/events/tests-EventController.cpp
```diff
@@ -0,0 +1,29 @@
...
+            add_to_the_global_ptr);
+
+        WHEN("an event is queued") {
+            std::shared_ptr<TestEvent> test_event = std::shared_ptr<TestEvent>(new TestEvent(1));
+            EventController::getInstance().queueEvent(test_event);
+
+            WHEN("events are processed") {
+                EventController::getInstance().forceProcessEvents();
+                REQUIRE(THE_GLOBAL == 1);
```
copacetic: nice test!

copacetic:
left some comments, looks good to me overall!

src/common/events/EventController.cpp
SourenP: Hence the "force" prefix. 
This means that the thread that it's up to the thread calling this function to make sure the used variables are thread safe to write/read from.

It's designed this way so that GameController (the master thread) can unlock/lock `EventController` in `collectClientInputs`, which is not the same place where `processEvents` is called.

src/server/NetworkingServer.cpp
SourenP: It keeps track of how many copies of the pointer exist and deletes it when there are none left. This avoids "memory leaks and dangling Pointers" https://thispointer.com/learning-shared_ptr-part-1-usage-details/

copacetic: is there any downside?

PR #68, opened 2019-09-12
Update dir structure

SourenP:


copacetic:
can you add a much more detailed commit message please? I dont know what this does or why

PR #71, opened 2019-09-25
Client Interpolation (rewind and replay) and variable physics time step 

SourenP:
Changes
- Move shared files to common dir
- Create base GameController class that's inherited by ClientGameController and ServerGameController
- Delete Client and use ClientGameController instead
- Delete ClientGroup/Mine and use _real_ Groups and Mines in client instead
- Have group store player ids instead of player class pointers and pass this data over the network to client
- Complete Player/Group class applyUpdate and getUpdate functions
- Rewind and replay interpolation (see [networked_physics_2004](https://web.archive.org/web/20181107181430/https://gafferongames.com/post/networked_physics_2004/))
- Variable physics time step (see [fix_your_timestep](https://web.archive.org/web/20181107181440/https://gafferongames.com/post/fix_your_timestep/))


SourenP:
Read this: https://www.gabrielgambetta.com/entity-interpolation.html and this https://www.gabrielgambetta.com/lag-compensation.html
We are currently doing "Dead Reckoning", maybe we should do "Entity interpolation" in the future.

src/client/systems/ClientGameController.cpp
```diff
@@ -0,0 +1,168 @@
+#include <thread>
```
copacetic: did you use `git mv` to move these files? it destroys the git history for the file otherwise. pretty annoying

copacetic:
Can you add a test for interpolation?

PR #73, opened 2019-10-01
Setup cmake

SourenP:


src/client/systems/NetworkingClient.cpp
```diff
@@ -15,150 +15,218 @@
...
+int NetworkingClient::getPlayerId() const {
+    return mPlayerId_ta;
+}
 
-    if (!mIsRegistered) {
-        throw std::runtime_error("Failed to register. Exiting.");
-    }
+GameState NetworkingClient::getGameState() {
+     std::lock_guard<std::mutex> mGameState_guard(mGameState_lock);
+    mGameStateIsFresh_ta = false;
```
copacetic: what is _ta?

copacetic:
can you rebase when ready? definitely want to see a description of why all the work with mutexes happened to protect the thread state when shutting down

src/client/systems/NetworkingClient.cpp
SourenP: I added a description in the README.txt about this.

SourenP:
> can you rebase when ready? definitely want to see a description of why all the work with mutexes happened to protect the thread state when shutting down

`518da75` Rewrite NetworkingClient and NetworkingServer to use explicit mutex locks+guards instead of using custom written thread safe classes.

PR #77, opened 2019-10-09
Networking thread improvements

SourenP:
- Add sleeps to threads to limit cpu usage
- Broadcast game state to all clients constantly instead of having
clients ask for it
- Have server and client send udp ports to eachother over the initial
tcp registration communication
- Clean up networking code


copacetic:
@SourenP  a bad pattern I'm noticing in these changes is that you're saying what things do, but not why. The why is a lot more important

SourenP:
Good point.
I’ll be more mindful of that in future commits/PRs.

PR #125, opened 2019-12-02
Mining mechanism

### LINKED ISSUE #14: Mining Mechanism

SourenP:
**Description**

*Group mines full capacity of mine.*
*Notice player's resource count rising on top right and how mine turns white when empty*
mining_mechanism

**Fixes**

Fixes #14 

**Commits**

[8c7b1de] Mining mechanism
- Mines now have a resource type and count. Their color is set according to their resource type and they will turn white when their count reaches 0. They start with a count of 4 for now.
- On collision, if the circles are a group and a mine, for each player in the group a single resource will be moved from the mine to the player.

### DIFF
--- a/src/common/objects/Mine.cpp
+++ b/src/common/objects/Mine.cpp
@@ -18,14 +18,25 @@ sf::Packet& operator>>(sf::Packet& packet, MineUpdate& mine_update) {
            mine_update.y_pos >> mine_update.radius >> mine_update.shader_key;
 }
 
-Mine::Mine(uint32_t id, sf::Vector2f position, float size, sf::Color color, PhysicsController& pc,
-           ResourceStore& rs) :
+Mine::Mine(uint32_t id, sf::Vector2f position, float size, sf::Color color,
+           ResourceType resource_type, PhysicsController& pc, ResourceStore& rs) :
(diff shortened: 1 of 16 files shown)


src/common/objects/Mine.cpp
```diff
@@ -18,16 +18,27 @@ sf::Packet& operator>>(sf::Packet& packet, MineUpdate& mine_update) {
...
+    m_resourceType(resource_type) {
     setTexture(RenderingDef::TextureKey::mine_pattern);
 }
 
 Mine::~Mine() {
 }
 
+void Mine::draw(sf::RenderTarget& render_target) {
+    if (m_resourceCount == 0) {
+        setColor(sf::Color::White);
```
SourenP: Make this a constant

src/common/systems/MineController.cpp
```diff
@@ -24,6 +25,7 @@ uint32_t MineController::createMine() {
     Mine& new_mine = *m_mines[next_mine_index];
     uint32_t new_mine_id = new_mine.getId();
     new_mine.setActive(true);
+    m_resourceController.add(new_mine_id, new_mine.getResourceType(), 4);
```
SourenP: Make mine start resource count a constant

PR #128, opened 2019-12-04
Winning Mechanism

### LINKED ISSUE #39: Winning Mechanism

SourenP:
**Description**

Stop game and display winner when a the win condition is met.

*Game over screen*
Screen Shot 2019-12-06 at 7 00 07 PM

**Fixes**

Fixes #39 

**Commits**

[b55e451] Create game over condition
- `State` is now synced over the network. I've created "none", "playing"
and "game_over" state.
- For now server will set state to "game_over" when any of the player's
resources match the win condition (a constant). The client will close
when state is "game_over".

[c35fbdc] Display winner player id at end of game
- Handle each game status step (pre/post/update) and draw accordingly
- Split game state into "core" and "object". Core contains meta data
about the game whereas object contains all the game object data.
- Replace (un)pack_game_state with packet operator overrides.

[3a56729] Minor fixes
- Rename remaining gameCoreState to gameStatCore
- Minor fixes

[3c8721b] Center game over winner text

[120ef2f] Split logic according to game status in server

### DIFF
--- a/src/common/util/game_def.hpp
+++ b/src/common/util/game_def.hpp
@@ -3,12 +3,6 @@
 
 #include <SFML/Graphics.hpp>
 
-enum State {
-    none,
-    playing,
-    game_over,
-};
-
 struct UnreliableCommand {
     sf::Uint32 client_id;
     sf::Uint32 command;
--- a/src/common/util/game_def.hpp
+++ b/src/common/util/game_def.hpp
@@ -3,6 +3,12 @@
 
 #include <SFML/Graphics.hpp>
 
+enum State {
+    none,
+    playing,
+    game_over,
+};
+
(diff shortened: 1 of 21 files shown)


src/common/util/game_def.hpp
```diff
@@ -3,6 +3,12 @@
 
 #include <SFML/Graphics.hpp>
 
+enum State {
```
copacetic: everything about this change looks fine, but please please don't call this thing State. It's the equivalent of calling it Thing. 

Also none should be probably be not_started right? If we frame it that way we can call it GameStatus or something

SourenP: Changed it to GameStatus

PR #136, opened 2019-12-09
Bots 0.01

SourenP:
**Description**

Adds bot functionality, headless mode, and command-line option support

Bot functionality: supports Random and NearestGreedy currently. Matches this spec:
https://docs.google.com/document/d/1Ew_wxM_rC5eHwUzXX3fQSAwvn7jSfd_9uGqy3eCJMag/edit?usp=sharing

Headless mode still opens a window unfortunately, so it's only a partial implementation. It still makes the client much lighter since all drawing is skipped. This is good because it means you can run many more clients locally when play testing.

Command line option support was added to toggle bot mode, headless mode and to choose the strategy you want the bots to follow. We will likely expand these options to specify server IP addresses and more down the line. The library I chose is called cxxopts.

**Fixes**

Nothing


src/client/bots/GreedyBot.cpp
```diff
@@ -0,0 +1,25 @@
...
+    };
+    m_unreliableInput = {
+        .toggle_up = false,
+        .toggle_down = false,
+        .toggle_right = bool(rand() % 2),
+        .toggle_left = false,
+        .toggle_stop = false,
+    };
+
+    return std::pair<InputDef::ReliableInput, InputDef::UnreliableInput>(m_reliableInput,
```
SourenP: Maybe the bot logic could produce a struct of more intuitive inputs and then we can have a util function all bots share that converts this to to `ReliableInput`/`UnreliableInput`.

src/client/bots/GreedyBot.hpp
```diff
@@ -0,0 +1,23 @@
+#ifndef GreedyBot_hpp
+#define GreedyBot_hpp
+
+#include "../../common/util/InputDef.hpp"
+
+#include <ctime>
+#include <cstdlib>
+
+class GreedyBot {
```
SourenP: Worth having a base bot class that this inherits from

src/client/systems/InputController.hpp
```diff
@@ -22,6 +28,8 @@ class InputController {
 
     InputDef::ReliableInput m_reliableInput;
     InputDef::UnreliableInput m_unreliableInput;
+
+    std::unique_ptr<GreedyBot> m_bot;
```
SourenP: This doesn't need to be a unique_ptr, it can just be allocated on the stack considering you're not passing it around or transferring ownership.
`GreedyBot m_bot`

src/common/util/game_settings.hpp
```diff
@@ -25,6 +25,8 @@ const sf::Vector2f WINDOW_RESOLUTION(2688, 1512);
 const float GAME_SCALE = 1.f;
 const sf::Color BACKGROUND_COLOR(sf::Color::Black);
 
+const bool IS_BOT = false;
```
SourenP: I don't think this should be defined here.
This is a constants file that's shared between the client and the server

Maybe we can have a class that sits between `InputController` and `ClientGameController` that's responsible for determining input and that class can be initialized as a bot or a human. 🤷‍♂

copacetic: Yep, my plan here is to actually take the "mode" from the command line or a config file. so you would run `ug-client --mode=bot --strategy=greedy` something like that

src/client/systems/InputController.hpp
copacetic: nice didn't know that, just copy pasta'd input controller

src/client/bots/GreedyBot.hpp
copacetic: yeah I had this fancy idea of strategy's and a bot which comes to life when you give it one, and even an idea of "tactics" which override strategies and come into effect in combat scenarios for example, but Im gonna just implement the different bots, see if there's shared functionality a base class could help with then refactor if it seems meaningful

src/client/bots/GreedyBot.cpp
copacetic: it does feel like this is a violation of the abstraction. like a bot shouldn't care about the reliable/unreliability of the input, just cares about a "move"

not high pri to fix for now, but worth keeping in mind yeah

src/client/systems/ClientGameController.cpp
```diff
@@ -48,7 +48,14 @@ InputDef::PlayerInputs ClientGameController::getPlayerInputs() {
...
 
 void ClientGameController::preUpdate() {
+    // call collectinputs to process window events, but if it's bot mode
+    // then override any user commands
     auto inputs = m_inputController->collectInputs(m_window);
+    if (IS_BOT) {
+        inputs = m_gameObjectController->getBotMove(m_playerId);
+    }
+    sendInputs(inputs);
+    saveInputs(inputs);
```
SourenP: Why do you have these here? 
They are called bellow in line 68-69 while the game status is `playing`. No reason to send/save moves before the game started.

src/common/systems/GameObjectController.cpp
```diff
@@ -145,4 +145,26 @@ std::pair<bool, uint32_t> GameObjectController::getGameOver() {
...
+        mine_id_to_resource_count[mine_id] = m_resourceController.get(
+                mine_id, mine->getResourceType());
+    }
+
+    // required for greedy bot and most map aware bots:
+    // resources
+    // current location
+    // mine locations
+    return m_bot.getMove(BotStrategy::NearestGreedy, mines,
+            current_resources, bot_group_center, mine_id_to_resource_count);
```
SourenP: Instead of passing all this data as parameters, it would be cleaner if you made a `BotData` struct and put it there. 
I did something similar for `UIData`.

Do you think `BotData` should be a union of the data needed by all bots or should each `BotStrategy` have it's won struct?

src/common/bots/Bot.cpp
```diff
@@ -0,0 +1,110 @@
...
+    srand(time(0));
+    bool up = bool(rand() % 2);
+    bool right = bool(rand() % 2);
+    InputDef::UnreliableInput unreliable_input = {
+        .toggle_up = up,
+        .toggle_down = !up,
+        .toggle_right = right,
+        .toggle_left = !right,
+        .toggle_stop = false,
+    };
```
SourenP: If I'm not mistaken, this bot can only go in diagonal directions...
Logic is off, you should be able to have `toggle_up` be true and everything else be false for example.

SourenP: Stopping is also a random behavior 
If players can stand still, mr rando 🤖  should be able to stand still too ;)

src/common/bots/Bot.cpp
```diff
@@ -0,0 +1,110 @@
...
+Bot::getNearestGreedyMove(std::vector<std::shared_ptr<Mine>> mines,
+                std::array<uint32_t, RESOURCE_TYPE_COUNT> current_resources,
+                sf::Vector2f bot_center,
+                std::unordered_map<uint32_t, uint32_t> mine_id_to_resource_count
+                ) {
+
+
+    // how expensive is this? We really only need to do it when we change direction, since
+    // if we know we've been moving towards a mine we can assume we're continuing to move towards it.
+    // in which case we can't possibly have gotten closer to a different mine.
```
SourenP: yeah keeping state that indicates whether you already have a goal (mine to move towards) will save you a lot of computation here. The goal can reset to None when you satisfy it or something

SourenP: (n log n) * calculating vector distance 
https://stackoverflow.com/questions/4484900/what-is-the-time-complexity-of-stdsort-in-the-c-standard-library

src/common/bots/Bot.cpp
```diff
@@ -0,0 +1,110 @@
+#include "Bot.hpp"
+
+Bot::Bot() {};
+
+std::pair<InputDef::ReliableInput, InputDef::UnreliableInput>
+    Bot::vectorToNearestMoveTransform(sf::Vector2f v) {
```
SourenP: I would move this to `common/util/Util.hpp`
We have the opposite function there that converts inputs into a vector (used by the server)

src/common/bots/Bot.cpp
```diff
@@ -0,0 +1,110 @@
...
+        sf::Vector2f(-1.f,-1.f), //left-up
+        sf::Vector2f(1.f,1.f),   //right-down
+        sf::Vector2f(-1.f,1.f),  //left-down
+        sf::Vector2f(1.f,-1.f),  //right-up
+    };
+    sf::Vector2f closestAllowedDirection = *std::max(directions.begin(), directions.end(),
+        [v](sf::Vector2f* dir_a, sf::Vector2f* dir_b) {
+            return VectorUtil::dot(v, *dir_a) > VectorUtil::dot(v, *dir_b);
+        }
+    );
```
SourenP: nice maths

src/client/systems/ClientGameController.cpp
```diff
@@ -51,7 +51,11 @@ void ClientGameController::preUpdate() {
     // call collectinputs to process window events, but if it's bot mode
     // then override any user commands
     auto inputs = m_inputController->collectInputs(m_window);
-    if (IS_BOT) {
+    // very annoying bug fixed by this % 10. When it wasn't present bot would
+    // enter a move every single time this function was called. It was buffering up
+    // to the point of being extremely dated so the bot would run off the map replaying
+    // old moves
```
SourenP: Hmm, i wonder if there is a more elegant solution

src/client/systems/ClientGameController.cpp
copacetic: nice catch bad copy pasta

src/common/bots/Bot.cpp
```diff
@@ -18,23 +19,28 @@ std::pair<InputDef::ReliableInput, InputDef::UnreliableInput>
         sf::Vector2f(-1.f,1.f),  //left-down
         sf::Vector2f(1.f,-1.f),  //right-up
     };
-    sf::Vector2f closestAllowedDirection = *std::max(directions.begin(), directions.end(),
-        [v](sf::Vector2f* dir_a, sf::Vector2f* dir_b) {
-            return VectorUtil::dot(v, *dir_a) > VectorUtil::dot(v, *dir_b);
+    std::sort(directions.begin(), directions.end(),
+        [v](sf::Vector2f dir_a, sf::Vector2f dir_b) {
+//			std::cout << "cos sim a,v" << VectorUtil::cos_sim(dir_a, v) << " " << dir_a.x << dir_a.y << std::endl;
+//			std::cout << "cos sim b,v" << VectorUtil::cos_sim(dir_b, v) << " " << dir_b.x << dir_b.y << std::endl;
```
SourenP: forgot to delete

SourenP: here and elsewhere

src/common/bots/Bot.cpp
copacetic: I saw that, but making it InputDef aware felt ugly

src/common/physics/VectorUtil.cpp
```diff
@@ -41,6 +41,27 @@ float VectorUtil::dot(const sf::Vector2f& a, const sf::Vector2f& b) {
...
+	denom_b += B.x * B.x;
+
+	dot += A.y * B.y;
+	denom_a += A.y * A.y;
+	denom_b += B.y * B.y;
+
+    if (denom_a == 0 || denom_b == 0)
+    {
+		std::cout << "cosine similarity is not defined whenever one or both "
+                "input vectors are zero-vectors." << std::endl;
```
SourenP: did you mean to return here or crash?

src/common/bots/Bot.cpp
copacetic: yeah Ive been thinking about his, probably won't optimize given N is just 4 here and it's not a noticeable issue.

src/common/bots/Bot.cpp
copacetic: good point on mr rando, he definitely didnt get the love he deserves

src/common/physics/VectorUtil.cpp
```diff
@@ -41,6 +41,27 @@ float VectorUtil::dot(const sf::Vector2f& a, const sf::Vector2f& b) {
...
+	denom_a += A.x * A.x;
+	denom_b += B.x * B.x;
+
+	dot += A.y * B.y;
+	denom_a += A.y * A.y;
+	denom_b += B.y * B.y;
+
+    if (denom_a == 0 || denom_b == 0)
+    {
+		std::cout << "cosine similarity is not defined whenever one or both "
```
copacetic: fix weird indentation in this function

src/common/systems/GameObjectController.cpp
copacetic: Im fine with it being a union unless we find it's annoying. What's the actual benefit of putting it in a struct?

SourenP:
See codeclimate failures

src/common/systems/GameObjectController.cpp
SourenP: It's cleaner that way and someone can quickly see all the information needed for a bot.
I much prefer it.

src/common/bots/Bot.cpp
SourenP: yeah whatever

src/common/bots/Bot.cpp
SourenP: InputDef is just input constants - stuff that's related to inputs. nbd
the real concern is the dependency on networking but we can fix that later

src/common/systems/GameObjectController.cpp
copacetic: The tricky thing in this case is that the current inputs to this function are a union of what any bot strategy might desire. That means that if we struct it, we'd have to just pass bits and pieces down to strategies so they just have access to the parts they need. Personally I don't think putting it in a struct matters and it adds a layer of obfuscation when you try to understand what's in the struct

src/common/bots/Bot.cpp
copacetic: done

src/common/physics/VectorUtil.cpp
copacetic: I changed it to use cerr, but it'll throw a divide by 0 exception here and with the error message I think that's good enough

src/client/systems/ClientGameController.cpp
copacetic: this was a red herring. I got rid of the modulus and it runs fine.

src/common/bots/Bot.cpp
copacetic: these are gone now

src/common/util/game_settings.hpp
```diff
@@ -34,10 +34,11 @@ const bool SHOW_DIRECTION_LINES = true;
 const bool SHOW_PLAYER_IDS = true;
 
 /* Game Logic */
+const bool IS_BOT = false;
```
SourenP: you don't need this anymore

src/client/CMakeLists.txt
```diff
@@ -17,4 +17,6 @@ add_library(client-lib
     rendering/RenderingController.cpp
 )
 
+add_subdirectory("extern/cxxopts")
```
SourenP: extern files shouldn't be in `src`. `src` is only for code we've written 
Mote this to the top level `extern` dir (where catch is)

src/client/main.cpp
```diff
@@ -1,12 +1,34 @@
...
+            "b,bot", "Runs game using a bot.")(
+            "s,strategy", "Runs game using specified strategy. Options are Random, NearestGreedy.",
+            cxxopts::value<int>());
+        auto result = options.parse(argc, argv);
+        bool is_headless = result["headless"].as<bool>();
+        bool is_bot = result["bot"].as<bool>();
+        bool strategy_opt_present = result.count("strategy");
+        BotStrategy strategy = BotStrategy::Random;
+        if (strategy_opt_present) {
+            strategy = (BotStrategy)result["strategy"].as<int>();
```
SourenP: Always use `static_cast<T>` over `(T)`
https://github.com/lefticus/cppbestpractices/blob/c4fb1cf8f8f0018c583437f79a9c05f3d263e99d/04-Considering_Safety.md#use-c-style-cast-instead-of-c-style-cast

in this case `static_cast<BotStrategy>(result["strategy"].as<int>());`

src/client/systems/ClientGameController.cpp
```diff
@@ -11,7 +11,15 @@
...
+    m_networkingClient(new NetworkingClient()), m_inputController(new InputController(INPUT_KEYS)),
+    m_renderingController(
+        new RenderingController(m_window, *m_gameObjectController, *m_resourceStore)) {
+}
+
+ClientGameController::ClientGameController(bool is_headless, bool is_bot, BotStrategy strategy) :
+    m_headless(is_headless), m_isBot(is_bot), m_strategy(strategy), GameController(),
     m_window(sf::VideoMode(WINDOW_RESOLUTION.x, WINDOW_RESOLUTION.y), "Ungroup", sf::Style::Close),
     m_networkingClient(new NetworkingClient()), m_inputController(new InputController(INPUT_KEYS)),
     m_renderingController(
```
SourenP: I don't like the amount of code duplicated here just to support different parameters 

You can just provide default parameters and have one constructor:
```cpp
bool is_headless = false, bool is_bot = false, BotStrategy strategy = BotStrategy::Random
```

src/common/bots/Bot.cpp
```diff
@@ -0,0 +1,73 @@
...
+                          sf::Vector2f bot_center,
+                          std::unordered_map<uint32_t, uint32_t> mine_id_to_resource_count) {
+    std::sort(mines.begin(), mines.end(),
+              [bot_center](std::shared_ptr<Mine> mine_a, std::shared_ptr<Mine> mine_b) {
+                  auto dist_a = VectorUtil::distance(bot_center, mine_a->getCenter());
+                  auto dist_b = VectorUtil::distance(bot_center, mine_b->getCenter());
+                  return dist_a < dist_b;
+              });
+
+    for (auto mine : mines) {
```
SourenP: rather than copying the shared_ptr here (costs cpu because you need to +1 the ptr count), I would take a reference in the loop by doing 
```
for(auto& mine : mines)
```

src/common/bots/Bot.cpp
```diff
@@ -0,0 +1,73 @@
+#include "Bot.hpp"
+
+Bot::Bot(){};
+
+/* This strategy will move towards the closest mine which has a resource the bot needs to win.
```
SourenP: docstring comment style is 
```
/**
 * stuff
 */
```

:P

src/common/bots/Bot.cpp
```diff
@@ -0,0 +1,73 @@
...
+    // if there's no point trying to mine, then move randomly.
+    return getRandomMove();
+}
+
+std::pair<InputDef::ReliableInput, InputDef::UnreliableInput> Bot::getRandomMove() {
+    InputDef::ReliableInput reliable_input;
+    reliable_input.setAll(false);
+    /*  When we run many random bots, if we seed them with just the time
+        then they'll all generate the same moves. XORing with the pid is
+        an easy trick for making them have different per-process seeds. */
```
SourenP: Please use `//` for inline (non docstring) comments

src/common/bots/Bot.cpp
```diff
@@ -0,0 +1,73 @@
...
+    srand(time(0) ^ getpid());
+    bool up = bool(rand() % 2);
+    bool down = !up ? bool(rand() % 2) : false;
+    bool right = bool(rand() % 2);
+    bool left = !up ? bool(rand() % 2) : false;
+    bool stop = bool(rand() % 2);
```
SourenP: sorry if I'm being nit picky but we don't need to use mods and if statements to generate random bools
```cpp
auto bool_gen = std::bind(std::uniform_int_distribution<>(0,1),std::default_random_engine());
InputDef::UnreliableInput unreliable_input = {
    .toggle_up = bool_gen(),
    .toggle_down = bool_gen(),
    .toggle_right = bool_gen(),
    .toggle_left = bool_gen(),
    .toggle_stop = bool_gen(),
};
```
see https://stackoverflow.com/questions/43329352/generating-random-boolean/43329456

src/common/bots/Bot.hpp
```diff
@@ -0,0 +1,37 @@
...
+enum BotStrategy { Random, NearestGreedy, MostNeededGreedy, Groupie, GreedieGroupie };
+
+class Bot {
+  public:
+    Bot();
+    ~Bot(){};
+    Bot(const Bot& temp_obj) = delete;
+    Bot& operator=(const Bot& temp_obj) = delete;
+
+    std::pair<InputDef::ReliableInput, InputDef::UnreliableInput>
```
SourenP: maybe we can make a typedef for `std::pair<InputDef::ReliableInput, InputDef::UnreliableInput>` in `InputDef.hpp`

src/common/util/Util.hpp
```diff
@@ -1,31 +1,14 @@
...
-        }
-        if (toggle_left) {
-            direction += sf::Vector2f(-1.f, 0.f);
-        }
-        if (toggle_right) {
-            direction += sf::Vector2f(1.f, 0.f);
-        }
-        return VectorUtil::normalize(direction);
-    };
+                                  bool toggle_left, bool toggle_stop);
```
SourenP: Can we move these to `InputUtil.hpp` and delete this file? 
Not a big fan of generalized util files where a bunch of random misc stuff pile up

SourenP:
Here is the patch that fixes the cxxopts: https://gist.github.com/SourenP/d7b80cffd625fe65e74434f36ff607b9

I'm unable to commit to this branch since I'm not a contributor to your fork

SourenP:
Created a PR to your branch to fix the cxxopts include: https://github.com/copacetic/ungroup_sfml/pull/1

src/common/util/game_settings.hpp
copacetic: good catch

src/client/main.cpp
copacetic: good call didnt know about that

src/client/systems/ClientGameController.cpp
copacetic: I didn't know C++ had default params, good call

src/common/bots/Bot.cpp
copacetic: this is what clang format does though :0

src/common/bots/Bot.cpp
copacetic: clang format does this

copacetic: and I prefer it for multiline comments anyway

src/common/bots/Bot.cpp
copacetic: The way you're doing it will produce contradictions and undefined behavior. In theory it could result in all trues.

src/common/bots/Bot.hpp
copacetic: Im  actually wary of adding abstraction just because we can. it means one more lookup everytime you see "Inputs" or w/e you end up calling it

src/common/util/Util.hpp
copacetic: I was surprised you wanted me to add to this file

copacetic: done

copacetic:
I made almost all the changes you mentioned. the ones I didn't I left comments with explanation, we can always revisit them later. Im going to merge today to move on to other things :)

PR #137, opened 2019-12-11
Update run script add build script

copacetic:
**Description**

Adds a build script which supports both Ninja and CMake directly, then switches run.sh to use the new build script.


copacetic:
self merging so my other branch can use this. minor change and we can always address it if you feel it should be different.

PR #145, opened 2019-12-17
Game bounds 

### LINKED ISSUE #142: Create map boundaries
Groups can now drift off the map. We should create map boundaries that groups can't cross.

SourenP:
**Description**

Create game bounds that cannot be left
game_borders

**Fixes**

Fixes #142

**Commits**

[560897f] Add game bounds as a big circle
- Movable circles (groups) can't leave the circle's bounds
- PhysicsController is in charge of enforcing and correcting bounds
- Draw game circle as different color from out of bounds

[8235c7f] Create mines in a ring around the map center
- Change a couple of game colors for fun

[9c9614a] Minor fix

[9cf974c] Minor cleanup

[baf6617] Add documentation


SourenP:
When a circle leaves the bounds, it gets pushed back towards the direction of the center of the map until it is entirely within bounds. This creates a weird effect where pushing towards the bounds makes you slide against it instead of just stopping.

I asked Pegor to do the geometry required to move the circle back in the direction it was moving when it left the bounds until it's fully within bounds, he responded with this solution:

> If R is the radius of the large circle and p0 is its center, and r is the radius of the small circle, and L is the line corresponding the the direction vector, then the center of the small circle would be the intersection of L with a circle of radius R-r and center p0.

image

> http://mathworld.wolfram.com/Circle-LineIntersection.html

> of course this gives you two points of intersection. How would you figure out which one you want in a way that is computationally efficient?

image

> x1-x0 and y1-y0 should have the same signs, respectively as x2-x0 and y2-y0
> Or x1-x0 and y1-y0 should have the same signs, respectively as the X and Y of the vector

SourenP:
Tried to implement the solution above that Pegor solved but it was really buggy so I'm going to just use the original solution.

Here is what I had implemented for reference: 

```cpp
float sgn(float x) {
    return (x < 0) ? -1 : 1;
}

sf::Vector2f CollisionUtil::getBoundsCorrection(const CircleRigidBody& circle,
                                                sf::Vector2f bounds_center, float bounds_radius) {
    float r = circle.getRadius() - bounds_radius;
    sf::Vector2f p_1 = circle.getCenter() - bounds_center;
    sf::Vector2f p_2 = circle.getVelocity() - bounds_center;
    float d_x = p_1.x - p_2.x;
    float d_y = p_1.y - p_2.y;
    float d_r = sqrt((d_x * d_x) + (d_y * d_y));
    float D = p_1.x * p_2.y - p_2.x * p_1.y;

    float x = (D * d_y + sgn(d_y) * d_x * sqrt(r * r * d_r * d_r - D * D)) / (d_r * d_r);
    float y = (-D * d_x + abs(d_y) * sqrt(r * r * d_r * d_r - D * D)) / (d_r * d_r);

    if (sgn(x) != sgn(p_2.x)) {
        x *= -1;
    }

    if (sgn(y) != sgn(p_2.y)) {
        y *= -1;
    }

    return sf::Vector2f(x + bounds_center.x, y + bounds_center.y);
}
```

PR #149, opened 2019-12-26
Move all rendering code to client

### LINKED ISSUE #147: ug-server doesn't run in a windowless environment
### Explanation

ug-server is meant to not depend upon a windowing system, so that we can run it from Linux server OS's. When I attempted to do just that, I had a rude awakening: it failed to launch because game controller is common code between server and client and requires a window to operate.

The reason it requires a window is that it includes ResourceController and calls addTexture on group's/mine's. Although this is totally OK for the client to do, since the code is shared with the server, it becomes problematic in a window-less environment.

Currently there is no distinction between a displayable group and a "data-only" group which means that groups require a texture as part of their initialization, which is why game controller is stuck trying to initialize them the way they like (which requires a window.)

### Stack Trace

```
#0  0x00007ffff79f1625 in raise () from /lib64/libc.so.6
#1  0x00007ffff79da8d9 in abort () from /lib64/libc.so.6
#2  0x00007ffff7f4752b in sf::priv::OpenDisplay() [clone .cold] () from /lib64/libsfml-window.so.2.5
#3  0x00007ffff7f585dc in sf::priv::GlxContext::GlxContext(sf::priv::GlxContext*) () from /lib64/libsfml-window.so.2.5
#4  0x00007ffff7f4987f in sf::priv::GlContext::initResource() () from /lib64/libsfml-window.so.2.5
#5  0x00007ffff7f9f1f3 in sf::Texture::Texture() () from /lib64/libsfml-graphics.so.2.5
#6  0x0000000000636a00 in ResourceStore::addTexture (this=0x60d000000040, key=RenderingDef::collision, texture_path=<incomplete type>, repeated=false) at /root/ungroup_game/src/common/resources/ResourceStore.cpp:27
#7  0x000000000063620d in ResourceStore::load (this=0x60d000000040) at /root/ungroup_game/src/common/resources/ResourceStore.cpp:12
#8  0x0000000000635f3d in ResourceStore::ResourceStore (this=0x60d000000040) at /root/ungroup_game/src/common/resources/ResourceStore.cpp:8
#9  0x0000000000627228 in GameController::GameController (this=0x7fffffffe120) at /root/ungroup_game/src/common/systems/GameController.cpp:17
#10 0x00000000005865ff in ServerGameController::ServerGameController (this=0x7fffffffe120) at /root/ungroup_game/src/server/systems/ServerGameController.cpp:4
```

SourenP:
**Description**

Move all rendering code to client so that we can run the server in a remote server.
This is a better design because rather than the game logic/data code being mushed together with the rendering code, we now have separate rendering classes that _query_ the game data and render.

**Fixes**

Fixes #147

**Commits**

[b9f4ef7] Move all rendering code to client
- Create GameObjectRenderer owned by RenderingController which is
responsible for rendering game objects
- Create DrawableGroup and DrawableMine which are the visual
representations of groups and mines. They are updated with the
group/mine properties at the draw step- Create GameObjectRenderer owned by RenderingController which is
responsible for rendering game objects
- Create DrawableGroup and DrawableMine which are the visual
representations of groups and mines. They are updated with the
group/mine properties at the draw step

[f0b9523] DrawableGroup/Mine/Circle code cleanup
- Create base class DrawableCircle for DrawableGroup/Mine that provides
setTexture/Shader methods
- Remove voronoi_background shader (unused)

[c013478] Minor fix


SourenP:
Successfully started the server on the remote server  🎉 
```
[souren@li1968-26 ungroup_game]$ ./build/src/server/ug-server
Starting ungroup game server.
```

PR #153, opened 2020-01-02
Fix internet networking

### LINKED ISSUE #156: SERVER_IP hard-coded
Currently the server IP is hard-coded meaning we need to recompile everytime we want to use a different server. An example is localhost vs ungroup.dev. This is annoying.

copacetic:
**Description**

* Send regular no-op UDP packets for NAT traversal
* Add support for using dynamic client IP
* Add support for passing server IP from client cmd line

**Fixes**

Fixes #156, #155, #154

### DIFF
--- a/src/client/systems/NetworkingClient.cpp
+++ b/src/client/systems/NetworkingClient.cpp
@@ -12,11 +12,12 @@
 #include "../../common/util/game_settings.hpp"
 #include "../../common/util/network_util.hpp"
 
-const sf::Time CLIENT_TCP_TIMEOUT = sf::milliseconds(100);
+const sf::Time CLIENT_TCP_TIMEOUT = sf::milliseconds(300);
 ;
 
-NetworkingClient::NetworkingClient() : m_gameState_t({}) {
-    std::cout << "Starting client..." << std::endl;
+NetworkingClient::NetworkingClient(const std::string& server_ip) :
+    m_gameState_t({}), m_serverIp(server_ip) {
+    std::cout << "Starting client." << std::endl;
 
     createTcpSocket(SERVER_TCP_PORT);
     createUdpSocket();
@@ -45,7 +46,8 @@ NetworkingClient::~NetworkingClient() {
 void NetworkingClient::createTcpSocket(unsigned short port) {
     std::lock_guard<std::mutex> m_tcpSocket_guard(m_tcpSocket_lock);
     m_tcpSocket_t = std::unique_ptr<sf::TcpSocket>(new sf::TcpSocket);
-    m_tcpSocket_t->connect(SERVER_IP, port);
+
+    m_tcpSocket_t->connect(sf::IpAddress(m_serverIp), port);
 }
 
 void NetworkingClient::createUdpSocket() {
@@ -129,7 +131,8 @@ void NetworkingClient::readRegistrationResponse() {
             m_tick_ta = static_cast<uint>(reliable_command.tick);
             registration_response >> m_serverUdpPort;
             std::cout << "Registered as client " << m_clientId_ta << " at tick " << m_tick_ta
-                      << std::endl;
+                      << std::endl
+                      << "Server UDP port is: " << m_serverUdpPort << std::endl;
             return;
         }
     }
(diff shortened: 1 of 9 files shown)


src/client/systems/NetworkingClient.cpp
```diff
@@ -260,9 +260,34 @@ void NetworkingClient::unreliableSend() {
     }
 }
 
+void NetworkingClient::sendNatPunch() {
```
copacetic: the UDP sending logic here might be possible to dedup but it's kind of annoying since the packets have different arguments here and in the Input command send

src/client/systems/NetworkingClient.cpp
```diff
@@ -12,7 +12,7 @@
 #include "../../common/util/game_settings.hpp"
 #include "../../common/util/network_util.hpp"
 
-const sf::Time CLIENT_TCP_TIMEOUT = sf::milliseconds(100);
+const sf::Time CLIENT_TCP_TIMEOUT = sf::milliseconds(300);
```
copacetic: this timeout might still be too low in a case where we have 500 ms latency, but we probably should just tell the player their connection is too fucked

src/client/systems/NetworkingClient.cpp
```diff
@@ -260,9 +260,34 @@ void NetworkingClient::unreliableSend() {
...
+    } else {
+        std::cout << "Failed to form packet" << std::endl;
+    }
+}
+
 void NetworkingClient::sendUnreliableInput() {
     std::lock_guard<std::mutex> m_unreliableInputs_guard(m_unreliableInputs_lock);
     if (m_unreliableInputs_t.empty()) {
+        srand(time(0));
+        bool send_nat_punch = (rand() % 10 == 0);
```
copacetic: we basically dont want to constantly send nat punches, but how do we decide when to send it? basically we should send it if we haven't sent an input command recently. this was my quick an dirty implementation which didnt involve keeping a timestamp for last sent input command, but I think Ill just switch to the timestamp solution

src/client/systems/NetworkingClient.cpp
```diff
@@ -274,6 +299,7 @@ void NetworkingClient::sendUnreliableInput() {
     UnreliableCommand unreliable_command = {(sf::Uint32)m_clientId_ta, unreliable_input_cmd,
                                             m_tick_ta};
 
+    std::cout << "we sending a UDP packet" << std::endl;
```
copacetic: delete

src/client/systems/NetworkingClient.cpp
```diff
@@ -260,9 +260,34 @@ void NetworkingClient::unreliableSend() {
     }
 }
 
+void NetworkingClient::sendNatPunch() {
+    sf::Packet packet;
+    sf::Uint32 unreliable_input_cmd = UnreliableCommandType::nat_punch;
+    UnreliableCommand unreliable_command = {(sf::Uint32)m_clientId_ta, unreliable_input_cmd,
+                                            m_tick_ta};
+
+    std::cout << "we sending a nat punch" << std::endl;
```
copacetic: delete

src/client/systems/NetworkingClient.cpp
copacetic: maybe the sending function will just take a packet that you want sent

src/common/util/game_settings.hpp
```diff
@@ -71,8 +71,8 @@ const std::chrono::milliseconds SERVER_BROADCAST_GAME_STATE_SLEEP(16);
 const std::chrono::milliseconds CLIENT_UNRELIABLE_RECV_SLEEP(SERVER_BROADCAST_GAME_STATE_SLEEP / 2);
 
 /* Networking */
-static const char* CLIENT_IP = "127.0.0.1";
+static const char* CLIENT_IP = "73.15.213.77";
```
copacetic: I dont want to commit this, so Im going to implement dynamically learning the client IP and using it before merging

src/client/systems/NetworkingClient.hpp
```diff
@@ -56,6 +58,10 @@ class NetworkingClient {
     // Sockets
     void createTcpSocket(unsigned short port);
     void createUdpSocket();
+    void sendUnreliable(sf::Packet packet);
+
+    time_t m_last_sent_nat_punch = time(0) - NAT_PUNCH_INTERVAL;
+    std::string m_server_ip;
```
SourenP: The style for member variables is camelcase - `m_serverIp` or `m_serverIP`

src/client/systems/NetworkingClient.hpp
```diff
@@ -56,6 +58,10 @@ class NetworkingClient {
     // Sockets
     void createTcpSocket(unsigned short port);
     void createUdpSocket();
+    void sendUnreliable(sf::Packet packet);
+
+    time_t m_last_sent_nat_punch = time(0) - NAT_PUNCH_INTERVAL;
```
SourenP: `m_lastSentNatPunch`

src/common/util/game_settings.hpp
```diff
@@ -38,6 +38,7 @@ const bool SHOW_PLAYER_IDS = true;
 /* Game Logic */
 const bool USE_INTERPOLATION_REPLAY = false;
 
+const int NAT_PUNCH_INTERVAL = 30;
```
SourenP: Do we really need to send this every 30ms? 
How do we know what to make this interval, what does it depend on? 

Would it work if we just did it once at the start of the connection and then never again?

SourenP: Oh just read your commit comment. 30sec sounds good

Btw I would use the sfml [clock](https://www.sfml-dev.org/documentation/2.5.1/classsf_1_1Clock.php) module instead of `<time.h>`. It has this nice feature where you can just get the elapsed time since the last time you called `clock.reset()` and its syntax makes the granularity of time very explicit e.g. `clock.getElapsedTime().asSeconds()`

copacetic: that's cool, I think I'll hold off on changing the code since it's fine and it's actually a different way of thinking about the problem to use a clock and elapsed time vs sampling the current time and keeping a timestamp. Rather spend my brainwaves on more core problems for Prototype 1

PR #161, opened 2020-01-06
Add run_bots ruby script

copacetic:
**Description**

Script makes it easier to spin up many bots in headless mode running
on random strategy. You can currently specify number of bots and the
server to point them to. 127.0.0.1 or ungroup.dev (takes hostnames).


SourenP:
please make a `/scripts` dir

copacetic:
I added a scripts dir and moved all top level scripts to it

PR #165, opened 2020-01-08
Improve metric collection and metrics collected 

### LINKED ISSUE #162: Improve debug metrics
SPS and UPS are opaque (even forgot what they stand for) and not very accurate because they averages over all of time making it hard to notice change. They should probably average over the last minute or something. 
We should also have metrics to track server and networking performance.

SourenP:
**Description**

- New TemporalMetric class that helps accurately track metrics over a span of time
- Track metrics FPS, UPS, XPS (network updates per second), TDA (tick delta average), TIK count
- Include Boost

Screen Shot 2020-01-09 at 5 25 06 PM

**Fixes**

Fixes #162


src/client/systems/ClientGameController.hpp
```diff
@@ -69,7 +69,8 @@ class ClientGameController : public GameController {
     uint32_t m_playerId = 0;
     bool m_headless, m_isBot;
     BotStrategy m_strategy;
-    std::string m_server_ip;
+    std::string m_serverIP;
+    uint m_applyServerStateCount = 0;
```
SourenP: delete this

src/common/systems/GameController.hpp
```diff
@@ -50,8 +53,9 @@ class GameController {
     sf::Clock m_clock;
     sf::Int32 m_elapsedTime = 0;
     sf::Int32 m_timeAccumulator = 0;
-    sf::Int32 m_stepCount = 0;
-    sf::Int32 m_updateCount = 0;
+
+    TemporalMetric m_stepMetric{120, sf::seconds(0.5f)};
+    TemporalMetric m_updateMetric{120, sf::seconds(0.5f)};
```
SourenP: rename these to be more specific
maybe `m_gameStepMetric` and `m_gameUpdateMetric `

src/client/systems/ClientGameController.cpp
```diff
@@ -71,14 +69,12 @@ void ClientGameController::preUpdate() {
         }
         case GameStatus::playing: {
             if (m_isBot) {
+                // Override user input with bot input
                 inputs = m_gameObjectController->getBotMove(m_playerId, m_strategy);
             }
             sendInputs(inputs);
             saveInputs(inputs);
-
-            if (m_networkingClient->getGameStateIsFresh()) {
```
copacetic: why delete this?

src/common/metrics/TemporalMetric.hpp
```diff
@@ -0,0 +1,57 @@
...
+ * This class currenly only supports uint counts.
+ */
+
+#include <SFML/System.hpp>
+#include <boost/circular_buffer.hpp>
+
+#ifndef TemporalMetric_hpp
+#define TemporalMetric_hpp
+
+class TemporalMetric {
```
copacetic: can you add some tests for this? It would suck if we got it wrong

SourenP: I was going to but then I realized how difficult it would be without mocking time.
Any ideas?

PR #174, opened 2020-01-21
Uniq win

### LINKED ISSUE #141: Unique win conditions for players
We currently have one condition that applies to all players.
We should create a unique and fair win condition for each player.

We need to be careful not to end up in a world state where it's impossible for a player to win. For example the scenario where all the red mines are depleted and a player needs red resources.
Perhaps to avoid this for the MVP we can make the mines have infinite (or a very large number) capacity.

copacetic:
**Description**

Adds unique win conditions for players. 
Each player has a "primary" resource of which they require the most and then secondary resource of which they need less.

*Resource goals shown in resource counts. Here blue is the primary resource.*
Screen Shot 2020-02-03 at 5 50 46 PM

**Fixes**

Fixes #141


src/common/objects/Player.hpp
```diff
@@ -69,6 +77,7 @@ class Player : public GameObject {
     bool m_ungroup = false;                    // If player wants to ungroup.
     ResourceType m_intent = ResourceType::RED; // Resource player intends to collect. This is
                                                // communicated to other players.
+    WinCondition m_win_condition;
```
SourenP: `s/m_win_condition/m_winCondition`
member variables use the style `m_camelCase`

src/common/systems/GameObjectController.cpp
```diff
@@ -125,7 +125,14 @@ GameObjectController::getPlayerResources(uint32_t player_id) {
...
     for (auto player_id : m_playerController.getActivePlayerIds()) {
-        if (m_resourceController.get(player_id) == WIN_CONDITION) {
+        auto player = m_playerController.getPlayer(player_id);
+        auto win_condition = player->getWinCondition();
+        std::cout << "WIN CONDITION";
+        for (size_t i = 0; i < RESOURCE_TYPE_COUNT; i++) {
+            std::cout << win_condition.getResourceCountsToWin()[i] << " ";
+        }
+        std::cout << std::endl;
+        if (player->getWinCondition().satisfiesCondition(m_resourceController.get(player_id))) {
```
SourenP: nice

src/common/systems/GameObjectController.cpp
```diff
@@ -125,7 +125,14 @@ GameObjectController::getPlayerResources(uint32_t player_id) {
 
 std::pair<bool, uint32_t> GameObjectController::getGameOver() {
     for (auto player_id : m_playerController.getActivePlayerIds()) {
-        if (m_resourceController.get(player_id) == WIN_CONDITION) {
+        auto player = m_playerController.getPlayer(player_id);
+        auto win_condition = player->getWinCondition();
+        std::cout << "WIN CONDITION";
```
SourenP: delete

src/common/systems/GameObjectController.cpp
```diff
@@ -125,7 +125,14 @@ GameObjectController::getPlayerResources(uint32_t player_id) {
...
+        for (size_t i = 0; i < RESOURCE_TYPE_COUNT; i++) {
+            std::cout << win_condition.getResourceCountsToWin()[i] << " ";
+        }
+        std::cout << std::endl;
```
SourenP: delete

src/server/systems/NetworkingServer.cpp
```diff
@@ -1,16 +1,4 @@
...
-#include <utility>
-#include <vector>
-
-#include "../../common/events/ClientConnectedEvent.hpp"
-#include "../../common/events/ClientDisconnectedEvent.hpp"
-#include "../../common/events/EventController.hpp"
-#include "../../common/events/PlayerCreatedEvent.hpp"
-#include "../../common/util/InputDef.hpp"
-#include "../../common/util/game_settings.hpp"
-#include "../../common/util/network_util.hpp"
```
SourenP: why :o

my intuition was to include these here in instead of the hpp because the hpp doesn't really use them

Maybe your change even makes this slower because the hpp file you moved these into is included somewhere else, it'll bring along all those dependencies (i dont know cpp well enough to be certain tho)
youknowhatimean?

src/common/models/WinCondition.cpp
```diff
@@ -0,0 +1,53 @@
+#include "WinCondition.hpp"
+
+WinCondition::~WinCondition() {}
+
+WinCondition::WinCondition() {
+    auto total_resource_requirements = WIN_CONDITION_RESOURCE_TOTAL;
+    srand(time(NULL));
+
+    // define amount of resources required for "primary" required resource
+    size_t quantityRequiredForPrimaryResource = (total_resource_requirements / RESOURCE_TYPE_COUNT) * 2;
```
SourenP: `s/quantityRequiredForPrimaryResource/quantity_required_for_primary_resource`
local variables use snake case

src/common/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ add_library(common-lib
     factories/IdFactory.cpp
     bots/Bot.cpp
     metrics/TemporalMetric.cpp
+    models/WinCondition.cpp
```
SourenP: so we consider this a model because it's an object with some state that can be queried (using its methods)?

src/common/models/WinCondition.cpp
```diff
@@ -0,0 +1,53 @@
...
+    total_resource_requirements -= quantityRequiredForPrimaryResource;
+    // pick which resource player will need the most of
+    size_t primaryResource = (rand() % RESOURCE_TYPE_COUNT);
```
SourenP: `s/primaryResource/primary_resource`
local variables use snake case

src/common/models/WinCondition.cpp
```diff
@@ -0,0 +1,53 @@
...
+        if (i == primaryResource) {
+            m_resourceCountsToWin[i] = quantityRequiredForPrimaryResource;
+        } else {
+            // make all other resources equal split (-1 is to exclude the primary resource)
+            m_resourceCountsToWin[i] = total_resource_requirements / (RESOURCE_TYPE_COUNT - 1);
+        }
+    }
+}
+
+bool WinCondition::satisfiesCondition(std::array<uint32_t, RESOURCE_TYPE_COUNT> currentResources) {
```
SourenP: `s/currentResources/current_resources`
local variables use snake case

SourenP: `satisfiesCondition` could use a test if you feel like writing one

src/server/systems/NetworkingServer.cpp
SourenP: Maybe revert 038960702e19489d2e88a68f118649797b838377
doesn't seem like you used that change

src/common/CMakeLists.txt
copacetic: models are basically just data structures for organizing data and associated functionality
vs a controller which might be organizing many different structures and be running a game loop etc
it's kind of a fuzzy definition, but this is a way of storing useful objects which might be referenced across controllers is my thinking. im mostly drawing on MVC paradigm here
https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller

PR #190, opened 2020-04-16
Chasing camera draft

SourenP:
**Description**

- Lerp camera position towards player instead of centering on player

**Fixes**

Fixes #X


SourenP:
The camera is making the player's group look very jittery and I'm not sure why

I'm not sure if this is the cause but seems like the `update_time` is not stable. 
We take two `MIN_TIME_STEP`s occasionally. Notice the `32`s from the [std::out](https://github.com/SourenP/ungroup_game/pull/190/files#diff-00b48b70d34f3103954f8ca97ac60645R78) of this change that logs `update_time`:
```
16
16
16
32
16
16
16
16
16
16
16
16
16
16
16
16
16
16
16
16
16
32
16
16
16
16
16
16
16
16
16
16
16
16
16
16
16
16
16
16
16
32
16
16
16
16
16
16
16
16
16
16
16
16
16
16
16
16
16
16
```

It's hard to see the jitter in this gif but figured I'd post anyways:
chasing_camera

PR #191, opened 2020-04-23
Update cmake to build on windows

SourenP:
**Description**

- Link dependencies to client-lib and server-lib before, in turn, linking them to their executables
  - See the path I took to discover this here: https://en.sfml-dev.org/forums/index.php?topic=27152.0
- Windows build having trouble with `uint` so I replaced them all with `uint32_t`
- Add cmake var `GUNGROUP_STATIC` to statically link sfml and gcc stuff
- Remove redundant `-std=c++11 ` from `CMAKE_CXX_FLAGS `
- Downgrade boost requirement to 1.55 because I couldn't find 1.65 on windows
- Handle `-pthreads` in a generic way with CMake's `find_package(Threads)`.

**Fixes**

Fixes #178

PR #193, opened 2020-04-25
Add terminal UI to server

copacetic:
* uses termbox library (added in this commit)
* adds two new "components" TermboxTextOutputArea and TermboxTable
* uses component to show a general output area, a table for player
  stats, and a table for server stats
* player stats table shows current direction, whether joinable, updates
  per second, player ID, and average drift
* server stats table shows steps per second, updates per second, and
  game state broadcasts per second

**Description**

Description of changes

**Fixes**

Fixes #X


extern/termbox/.gitignore
```diff
@@ -0,0 +1,5 @@
+build
```
copacetic, line 1: delete?

src/server/systems/NetworkingServer.cpp
```diff
@@ -39,6 +37,14 @@ NetworkingServer::~NetworkingServer() {
         m_stateUdpSocket_t->unbind();
     }
 
+    // disconnect any connected clients
+    // if you don't do this you will get the "failed to bind to socket"
+    // error message when restarting the server.
```
copacetic, line 42: fun fact: this still happens

src/server/systems/NetworkingServer.cpp
```diff
@@ -168,23 +174,41 @@ void NetworkingServer::handleUnreliableCommand(sf::Socket::Status status, sf::Pa
             break;
         }
         default: {
-            std::cout << "Unknown command code sent: " << unreliable_command.command << std::endl;
+            std::cerr << "Unknown command code sent: " << unreliable_command.command << std::endl;
```
copacetic, line 177: currently cerr clobbers terminal UI, redirecting stderr to a file should solve this issue, but Im brainstorming ways to output cerr to the termbox terminal UI (probably means using a signleton)

copacetic: but im worried about thread safety issues with the singleton

src/server/systems/NetworkingServer.cpp
```diff
@@ -168,23 +174,41 @@ void NetworkingServer::handleUnreliableCommand(sf::Socket::Status status, sf::Pa
...
+            return;
+        }
+        player_id = m_clientToPlayerIds_t[client_id];
+    }
+
     int drift = std::abs(static_cast<int>((m_tick_ta - client_tick)));
+    {
+        std::lock_guard<std::mutex> m_playerIdsToDriftMetric_guard(m_playerIdsToDriftMetrics_lock);
+        if (m_playerIdsToDriftMetrics_t.find(player_id) == m_playerIdsToDriftMetrics_t.end()) {
+            m_playerIdsToDriftMetrics_t[player_id] = new TemporalMetric{120, sf::seconds(0.5f)};
```
copacetic, line 198: I should switch this to 30 like the rest of these

src/server/systems/NetworkingServer.cpp
```diff
@@ -168,23 +174,41 @@ void NetworkingServer::handleUnreliableCommand(sf::Socket::Status status, sf::Pa
...
+        m_playerIdsToDriftMetrics_t[player_id]->update();
+
+    }
     if (drift < CMD_DRIFT_THRESHOLD) {
-        sf::Uint32 player_id;
         {
-            std::lock_guard<std::mutex> m_clientToPlayerIds_guard(m_clientToPlayerIds_lock);
-            if (m_clientToPlayerIds_t.count(client_id) == 0) {
-                return;
+            std::lock_guard<std::mutex> m_playerIdsToUpdatesMetric_guard(m_playerIdsToUpdatesMetric_lock);
```
copacetic, line 206: code is duplicated but only once and it's clear what it does vs deduping might make it less clear, leaning towards leaving as is until we have more code like this

src/server/systems/NetworkingServer.cpp
```diff
@@ -194,9 +218,38 @@ void NetworkingServer::setClientUnreliableUpdate(sf::Packet packet, int client_i
                 m_playerUnreliableUpdates_lock);
             m_playerUnreliableUpdates_t.push_back(player_unreliable_input);
         }
-    } else {
-        std::cout << "Receive client_update command with tick drifted past drift threshold. "
-                  << "Drift value is: " << drift << std::endl;
+    }
+}
+
```
copacetic, line 223: extra newline

src/server/systems/NetworkingServer.cpp
```diff
@@ -194,9 +218,38 @@ void NetworkingServer::setClientUnreliableUpdate(sf::Packet packet, int client_i
...
+        return m_playerIdsToUpdatesRates_t;
+    }
+}
+
+std::unordered_map<sf::Uint32, float> NetworkingServer::getPlayerTickDrifts() {
+    {
+        std::lock_guard<std::mutex> m_playerIdsToDriftMetric_guard(m_playerIdsToDriftMetrics_lock);
+        auto map = m_playerIdsToDriftMetrics_t;
+        for (auto it = map.begin(); it != map.end(); ++it) {
+            m_playerIdsToDrifts_t[it->first] = (it->second)->getAverage();//Rate(sf::seconds(1));
```
copacetic, line 249: delete comment

src/server/systems/NetworkingServer.hpp
```diff
@@ -94,6 +98,17 @@ class NetworkingServer {
...
+    std::mutex m_broadcastGameStateMetric_lock;;
+    TemporalMetric m_broadcastGameStateMetric_t{30, sf::seconds(0.5f)};
+
+    std::mutex m_playerIdsToUpdatesMetric_lock;
+    std::unordered_map<sf::Uint32, TemporalMetric*> m_playerIdsToUpdatesMetric_t;
+    std::unordered_map<sf::Uint32, float> m_playerIdsToUpdatesRates_t;
+
+    std::mutex m_playerIdsToDriftMetrics_lock;
+    std::unordered_map<sf::Uint32, TemporalMetric*> m_playerIdsToDriftMetrics_t;
+    std::unordered_map<sf::Uint32, float> m_playerIdsToDrifts_t;
```
copacetic, line 110: I compute into this float unordered_map which is then returned to ServerGameController (and different thread) to avoid thread safety issues on the TemporalMetric objects

src/server/systems/ServerGameController.cpp
```diff
@@ -2,15 +2,107 @@
...
 
+void ServerGameController::draw() {
+    auto playerIdsToUpdatesRates = m_networkingServer->getPlayerUnreliableUpdatesRates();
+    auto playerIdsToAvgTickDrifts = m_networkingServer->getPlayerTickDrifts();
+
+    auto broadcastGameStateRate = m_networkingServer->getBroadcastGameStateRate();
+
+    GameStateObject gso = m_gameObjectController->getGameStateObject();
+    auto player_updates = gso.player_updates;
+    auto total_active_players = std::to_string(
```
copacetic, line 34: delete, unused

src/server/systems/ServerGameController.cpp
```diff
@@ -2,15 +2,107 @@
...
+            [](PlayerUpdate pu){return pu.is_active;}
+        )
+    );
+
+    uint32_t curr_row = 1;
+    uint32_t curr_col;
+    // clear out internal termbox screen buffer
+    tb_clear();
+
+    m_termboxTextOutputArea.setMaxHeight(tb_height()/2);
```
copacetic, line 47: maybe add comments explaining why this is done (to create panes in the view)

src/server/util/TermboxTextOutputArea.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+void TermboxTextOutputArea::setMaxHeight(uint32_t max_height) {
+    if (max_height < MAX_LINES) {   
+        m_max_height = max_height;
+    } else {
+        std::cerr << "You must set a max height less than MAX_LINES" << std::endl;
+    }
+}
+
+uint32_t TermboxTextOutputArea::setStrAtCoord(std::string str, uint32_t x, uint32_t y, uint16_t fg) {
```
copacetic, line 33: maybe create termbox util since this function is shared with other termbox component

src/server/systems/ServerGameController.cpp
```diff
@@ -2,15 +2,107 @@
...
+    m_termboxPlayerTable.addColumn("Player ID");
+    m_termboxPlayerTable.addColumn("UPS");
+    m_termboxPlayerTable.addColumn("Avg Drift");
+    m_termboxPlayerTable.addColumn("Direction");
+    m_termboxPlayerTable.addColumn("Joinable");
+    m_termboxServerStatsTable.addColumn("Steps per second");
+    m_termboxServerStatsTable.addColumn("Updates per second");
+    m_termboxServerStatsTable.addColumn("Broadcasts per second");
+    m_termboxTextOutputArea.addLine("Ungroup Game Server Output");
+    m_termboxTextOutputArea.addLine("--------------------------");
```
SourenP, line 20: We should move all of this initialization code and the stuff in draw into its own class.
The class can have `update` `preUpdate` `postUpdate` and `draw` methods.

I think it's really important to keep the core game loop as simple and understandable as possible.



This is how it looks on the client side:

**init**
https://github.com/SourenP/ungroup_game/blob/b47a337b15ba2544f5986de6805269b170fe6bfa/src/client/systems/ClientGameController.cpp#L48-L49
**draw**
https://github.com/SourenP/ungroup_game/blob/b47a337b15ba2544f5986de6805269b170fe6bfa/src/client/systems/ClientGameController.cpp#L57-L59

src/common/systems/GameController.hpp
```diff
@@ -52,8 +52,8 @@ class GameController {
     sf::Int32 m_elapsedTime = 0;
     sf::Int32 m_timeAccumulator = 0;
 
-    TemporalMetric m_gameStepMetric{120, sf::seconds(0.5f)};
-    TemporalMetric m_gameUpdateMetric{120, sf::seconds(0.5f)};
+    TemporalMetric m_gameStepMetric{30, sf::seconds(0.5f)};
+    TemporalMetric m_gameUpdateMetric{30, sf::seconds(0.5f)};
```
SourenP, line 56: It would be nice to make these constants. 
+ i want to make a todo that surfaces this number to the UI so there is more transparency of how the metric was calculated. Im not too worried about it being verbose since it's for debugging anyways

src/server/systems/NetworkingServer.cpp
SourenP: can you change this comment then, its misleading

src/server/systems/NetworkingServer.cpp
```diff
@@ -194,9 +218,38 @@ void NetworkingServer::setClientUnreliableUpdate(sf::Packet packet, int client_i
...
+        std::lock_guard<std::mutex> m_playerIdsToUpdatesMetric_guard(m_playerIdsToUpdatesMetric_lock);
+        return m_broadcastGameStateMetric_t.getRate(sf::seconds(1));
+    }
+}
+
+std::unordered_map<sf::Uint32, float> NetworkingServer::getPlayerUnreliableUpdatesRates() {
+    {
+        std::lock_guard<std::mutex> m_playerIdsToUpdatesMetric_guard(m_playerIdsToUpdatesMetric_lock);
+        auto map = m_playerIdsToUpdatesMetric_t;
+        for (auto it = map.begin(); it != map.end(); ++it) {
```
SourenP, line 236: can you change these iterator loops to the range based style, it's much more readable 

```
for (const auto& kv : myMap) {
    std::cout << kv.first << " has value " << kv.second << std::endl;
}
```

src/server/systems/NetworkingServer.cpp
```diff
@@ -194,9 +218,38 @@ void NetworkingServer::setClientUnreliableUpdate(sf::Packet packet, int client_i
...
+
+        return m_playerIdsToUpdatesRates_t;
+    }
+}
+
+std::unordered_map<sf::Uint32, float> NetworkingServer::getPlayerTickDrifts() {
+    {
+        std::lock_guard<std::mutex> m_playerIdsToDriftMetric_guard(m_playerIdsToDriftMetrics_lock);
+        auto map = m_playerIdsToDriftMetrics_t;
+        for (auto it = map.begin(); it != map.end(); ++it) {
```
SourenP, line 248: can you change these iterator loops to the range based style, it's much more readable 

```
for (const auto& kv : myMap) {
    std::cout << kv.first << " has value " << kv.second << std::endl;
}
```

src/server/util/TermboxTextOutputArea.cpp
```diff
@@ -0,0 +1,39 @@
+#include "TermboxTextOutputArea.hpp"
```
SourenP, line 1: Not going to review this termbox layout code, i hope you don't mind

src/server/systems/NetworkingServer.cpp
```diff
@@ -194,9 +218,38 @@ void NetworkingServer::setClientUnreliableUpdate(sf::Packet packet, int client_i
...
+        }
+
+        return m_playerIdsToUpdatesRates_t;
+    }
+}
+
+std::unordered_map<sf::Uint32, float> NetworkingServer::getPlayerTickDrifts() {
+    {
+        std::lock_guard<std::mutex> m_playerIdsToDriftMetric_guard(m_playerIdsToDriftMetrics_lock);
+        auto map = m_playerIdsToDriftMetrics_t;
```
SourenP, line 247: You dont need to copy the map here since you're holding the lock for the entire scope anyways,
you can just loop over `m_playerIdsToDriftMetrics_t` and copy the data into `m_playerIdsToDrifts_t`

src/server/systems/NetworkingServer.cpp
```diff
@@ -168,23 +174,41 @@ void NetworkingServer::handleUnreliableCommand(sf::Socket::Status status, sf::Pa
...
     int drift = std::abs(static_cast<int>((m_tick_ta - client_tick)));
+    {
+        std::lock_guard<std::mutex> m_playerIdsToDriftMetric_guard(m_playerIdsToDriftMetrics_lock);
+        if (m_playerIdsToDriftMetrics_t.find(player_id) == m_playerIdsToDriftMetrics_t.end()) {
+            m_playerIdsToDriftMetrics_t[player_id] = new TemporalMetric{120, sf::seconds(0.5f)};
+        }
+        m_playerIdsToDriftMetrics_t[player_id]->pushCount(drift);
+        m_playerIdsToDriftMetrics_t[player_id]->update();
+
+    }
```
SourenP, line 203: maybe worth moving this to its own method, the networking code is already a bit tough to read so it would be nice if you could just read one line for all of this and think "oh and we log a drift metric here"

src/server/systems/NetworkingServer.cpp
```diff
@@ -194,9 +218,38 @@ void NetworkingServer::setClientUnreliableUpdate(sf::Packet packet, int client_i
...
+    {
+        std::lock_guard<std::mutex> m_playerIdsToUpdatesMetric_guard(m_playerIdsToUpdatesMetric_lock);
+        return m_broadcastGameStateMetric_t.getRate(sf::seconds(1));
+    }
+}
+
+std::unordered_map<sf::Uint32, float> NetworkingServer::getPlayerUnreliableUpdatesRates() {
+    {
+        std::lock_guard<std::mutex> m_playerIdsToUpdatesMetric_guard(m_playerIdsToUpdatesMetric_lock);
+        auto map = m_playerIdsToUpdatesMetric_t;
```
SourenP, line 235: You dont need to copy the map here since you're holding the lock for the entire scope anyways,
you can just loop over `m_playerIdsToUpdatesMetric_t` and copy the data into `m_playerIdsToUpdatesRates_t`

src/server/systems/NetworkingServer.cpp
```diff
@@ -168,23 +174,41 @@ void NetworkingServer::handleUnreliableCommand(sf::Socket::Status status, sf::Pa
...
+        if (m_clientToPlayerIds_t.count(client_id) == 0) {
+            return;
+        }
+        player_id = m_clientToPlayerIds_t[client_id];
+    }
+
     int drift = std::abs(static_cast<int>((m_tick_ta - client_tick)));
+    {
+        std::lock_guard<std::mutex> m_playerIdsToDriftMetric_guard(m_playerIdsToDriftMetrics_lock);
+        if (m_playerIdsToDriftMetrics_t.find(player_id) == m_playerIdsToDriftMetrics_t.end()) {
```
SourenP, line 197: you can replace this with `m_playerIdsToDriftMetrics_t.count(player_id) == 0`

SourenP:
Before you merge this, i'll run it through clang-format on my computer to fix the style

SourenP:
According to travis the smoke test failed. I think it's because it relies on stdout to print something specific

```
1.16s$ bash -c 'if ! xvfb-run --server-args="-screen 0 1280x1024x24" -e /dev/stdout -a node tests/integration/smoke_tests.js; then >&2 echo "Fail"; exit 1; fi'
eoke test failed!
Fail
The command "bash -c 'if ! xvfb-run --server-args="-screen 0 1280x1024x24" -e /dev/stdout -a node tests/integration/smoke_tests.js; then >&2 echo "Fail"; exit 1; fi'" exited with 1.
Done. Your build exited with 1.
```

SourenP:
I'll merge #191 after you merge this into master

src/server/util/TermboxTextOutputArea.cpp
copacetic: decided not to, not enough duplication to be worth the effort

copacetic:
🥳

PR #197, opened 2020-04-28
various server improvements

copacetic:
**Description**

*Add live tick to server UI*

*Add option to specify TCP port to listen on / connect to on server / client respectively*

*Show TCP port server is listening on in server UI*

*Add script to launch the server on an available TCP port, optionally opens firewall rules on Fedora*

### DIFF
--- a/src/client/main.cpp
+++ b/src/client/main.cpp
@@ -16,6 +16,7 @@ int main(int argc, char** argv) {
             "x,headless", "Runs game without updating window.")("b,bot", "Runs game using a bot.")(
             "a,server-ip", "The server's IP address in CIDR notation (ex. 127.0.0.1)",
             cxxopts::value<std::string>())(
+            "p,port", "The TCP port to connect to.", cxxopts::value<int>())(
             "s,strategy", "Runs game using specified strategy. Options are Random, NearestGreedy.",
             cxxopts::value<int>());
         auto result = options.parse(argc, argv);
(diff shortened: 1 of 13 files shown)


src/client/main.cpp
```diff
@@ -41,7 +42,12 @@ int main(int argc, char** argv) {
             strategy = static_cast<BotStrategy>(result["strategy"].as<int>());
         }
 
-        ClientGameController client_game_controller(is_headless, is_bot, strategy, server_ip);
+        bool tcp_port_opt_present = result.count("port");
+        uint32_t tcp_port = GAME_SETTINGS.SERVER_TCP_PORT;
+        if (tcp_port_opt_present) {
```
SourenP, line 47: I would just replace this with `result.count("port")` instead of using the extra var `tcp_port_opt_present`

src/server/main.cpp
```diff
@@ -28,7 +29,12 @@ int main(int argc, char** argv) {
             level = static_cast<LevelKey>(result["level"].as<int>());
         }
 
-        ServerGameController server_game_controller(level);
+        bool tcp_port_opt_present = result.count("port");
+        uint32_t tcp_port = GAME_SETTINGS.SERVER_TCP_PORT;
+        if (tcp_port_opt_present) {
```
SourenP, line 34: ditto
we dont need this var, put the `.count` straight in the if

src/server/rendering/TerminalRenderingController.cpp
```diff
@@ -1,6 +1,6 @@
 #include "TerminalRenderingController.hpp"
 
-TerminalRenderingController::TerminalRenderingController() : m_termboxTextOutputArea(10) {
+TerminalRenderingController::TerminalRenderingController(uint32_t tcp_port) : m_termboxTextOutputArea(10) {
```
SourenP, line 3: Why are you passing this as a constructor instead of setting it like the other values via draw?

src/server/main.cpp
copacetic: I think it's more readable naming hwy we're getting the count personally

src/server/rendering/TerminalRenderingController.cpp
copacetic: no real reason, doing it this way meant not saving a variable in ServerGameController. less code/noise

PR #207, opened 2020-05-01
Refactor bot code

copacetic:
Move into ClientGameController and simplify how data is passed in by
just passing in GameObjectController. Means I don't have to update
function definitions every time I want to pass any data...

Also, I fixed the greedy bot to actually take into account the win
conditions, instead of dumbly trying to consume _all_ resources.

**Description**

Description of changes

**Fixes**

Fixes #X

### DIFF
--- a/src/common/bots/Bot.cpp
+++ b/src/common/bots/Bot.cpp
@@ -6,10 +6,23 @@ Bot::Bot(){};
  * Once it has enough of a resource, it will move to the next closest mine with a resource it needs.
  */
 std::pair<InputDef::ReliableInput, InputDef::UnreliableInput>
-Bot::getNearestGreedyMove(std::vector<std::shared_ptr<Mine>> mines,
-                          std::array<uint32_t, RESOURCE_TYPE_COUNT> current_resources,
-                          sf::Vector2f bot_center,
-                          std::unordered_map<uint32_t, uint32_t> mine_id_to_resource_count) {
+Bot::getNearestGreedyMove(uint32_t bot_player_id, GameObjectController& goc) {
+    auto gso = goc.getGameStateObject();
+    auto bot_center = goc.getPlayerPosition(bot_player_id);
+    auto resource_controller = goc.getResourceController();
+    auto& player_controller = goc.getPlayerController();
+    auto current_resources = resource_controller.get(bot_player_id);
+    auto mines = (goc.getMineController()).getMines();
+
+    auto bot_player = player_controller.getPlayer(bot_player_id);
+    auto& bot_win_condition = bot_player->getWinCondition();
+    auto resource_counts_to_win = bot_win_condition.getResourceCountsToWin();
+    std::unordered_map<uint32_t, uint32_t> mine_id_to_resource_count;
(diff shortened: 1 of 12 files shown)


src/common/systems/MineController.hpp
```diff
@@ -16,6 +16,9 @@ class MineController {
     uint32_t createMine(sf::Vector2f center_position);
     void update();
     Mine& getMine(uint32_t mine_id);
+    std::vector<std::shared_ptr<Mine>>& getMines() {
```
SourenP, line 19: We dont need this method considering we have getMines in GameObjectController
https://github.com/SourenP/ungroup_game/blob/4e387552297a774f777d2d2a549148cbffc172d3/src/common/systems/GameObjectStore.cpp#L59

src/common/bots/Bot.cpp
```diff
@@ -6,10 +6,23 @@ Bot::Bot(){};
...
-                          std::array<uint32_t, RESOURCE_TYPE_COUNT> current_resources,
-                          sf::Vector2f bot_center,
-                          std::unordered_map<uint32_t, uint32_t> mine_id_to_resource_count) {
+Bot::getNearestGreedyMove(uint32_t bot_player_id, GameObjectController& goc) {
+    auto gso = goc.getGameStateObject();
+    auto bot_center = goc.getPlayerPosition(bot_player_id);
+    auto resource_controller = goc.getResourceController();
+    auto& player_controller = goc.getPlayerController();
+    auto current_resources = resource_controller.get(bot_player_id);
+    auto mines = (goc.getMineController()).getMines();
```
SourenP, line 15: call `goc.getMines()` instead

src/common/bots/Bot.cpp
```diff
@@ -6,10 +6,23 @@ Bot::Bot(){};
  * Once it has enough of a resource, it will move to the next closest mine with a resource it needs.
  */
 std::pair<InputDef::ReliableInput, InputDef::UnreliableInput>
-Bot::getNearestGreedyMove(std::vector<std::shared_ptr<Mine>> mines,
-                          std::array<uint32_t, RESOURCE_TYPE_COUNT> current_resources,
-                          sf::Vector2f bot_center,
-                          std::unordered_map<uint32_t, uint32_t> mine_id_to_resource_count) {
+Bot::getNearestGreedyMove(uint32_t bot_player_id, GameObjectController& goc) {
```
SourenP, line 9: crazy how much computation this does per frame

src/common/systems/MineController.hpp
copacetic: you linked GameObjectStore

src/common/bots/Bot.cpp
copacetic: it's in GameObjectStore not in the controller

src/common/bots/Bot.cpp
copacetic: yeah, it feels like the right decision that we did 1 process per bot for that reason. bots stills only take up 3.5% CPU on avg on my ultrabook

PR #209, opened 2020-05-02
Fade other players' directions in group

### LINKED ISSUE #208: Hard to tell players own direction in large group

SourenP:
**Description**

_Group with three players_
Screen Shot 2020-05-01 at 5 17 04 PM

**Fixes**

Fixes #208


copacetic:
does it render the arrow in the direction the group is moving or does it show where you are saying the group to go?

In the first case, the arrow will always map to the actual movement of the group, in the second it could show right but the group could be moving left if the two other players say left

SourenP:
> does it render the arrow in the direction the group is moving or does it show where you are saying the group to go?
> 
> In the first case, the arrow will always map to the actual movement of the group, in the second it could show right but the group could be moving left if the two other players say left

It shows the direction of each player, not the group. Your own direction arrow's color is full alpha whereas the others' is transparent.

The arrow that shows the direction of the group is currently hidden. Right now it feels like we don't need it, just adds another arrow that can be confusing.

PR #213, opened 2020-05-14
Play in the past

sourencho:
**Description**

With the server running on the remote server and the client connected to the ethernet with a ping of ~15ms the experience is fairly smooth. Changing direction isn't instant but the transition is smooth. The group still jitters (even when moving in the same direction) but it's very local, doesn't jump back and forth too far.

_ethernet ping and debug metrics_
Screen Shot 2020-05-23 at 1 22 26 PM

The game is also fairly smooth over wifi and consistent ~18ms ping 

_wifi ping and debug metrics_
Screen Shot 2020-05-23 at 2 39 50 PM

Changes:
- Replace rewind and replay with the client playing in the past and
interpolating to game states recieved from the server stored in a fixed
size buffer. The size of the buffer is configurable through the var
CLIENT_GAME_STATE_BUFFER_SIZE in game_settings.json.
- The logic for updating the local game state is as follows:
  - If the clients tick is too far behind (more than
  MAX_BEHIND_TICK_DELTA ticks) the smallest server state in the buffer
  then the local state is overwritten with the smallest server state.
  - If the client tick is not too far behind the smallest server state
  then then a game state is interpolated for the next tick using the
  smallest server state.
  - If the client is ahead of the smallest server state then the next
  game state in the buffer is used to recalculate the current tick's
  state but the tick is not incremented. This is in order to stall and
  fall back behind the smallest server state. If there aren't any states
  in the buffer ahead of the current tick, then the game state doesn't
  change.
- Add debug metrics around interpolation, ahead, behind, etc and update
the layout to be more readable
- Implement hermite interpolation to be able interpolate curves rather
than straight lines given start and end velocity of the rigid bodies.
- Send collision events in game state updates. This was necessary
because we no longer extrapolate, we only interpolate, which means we no
longer run the physics simulation locally and can't trigger visuals
caused by collisions. The events recieved from the server are now
applied and trigger collision animations. This introduced a visual bug
(issue #214).
- Reduce server input window sleep to 16ms to make the client and server
ticks/sec in sync. This is a messy dependency, it just so happens that
the server sleeps for that window causing it to control it's steps per
second. Not really a good way to control tick speed and the client and
server tick speeds goes out of sync as more clients connect. We might
need to implement some clock sync in the future.
- Seems like this change also broke the winner being displayed on the
client (issue #215).
- Minor code cleanup.

**Fixes**

Fixes #X

### DIFF
--- a/resources/configs/game_settings.json
+++ b/resources/configs/game_settings.json
@@ -10,14 +10,14 @@
         "TOTAL_RESOURCE_REQUIREMENTS": 20
     },
     "core": {
-        "USE_REMAINDER_DELTA_TIME": false
+        "USE_REMAINDER_DELTA_TIME": true
     },
     "physics": {
         "MIN_TIME_STEP_SEC": 0.008
     },
     "threads": {
         "CLIENT_FETCH_PLAYER_ID_SLEEP": 16,
-        "SERVER_INPUT_WINDOW_SLEEP": 50,
+        "SERVER_INPUT_WINDOW_SLEEP": 16,
         "CLIENT_RELIABLE_SEND_SLEEP": 16,
         "CLIENT_UNRELIABLE_SEND_SLEEP": 16,
         "SERVER_BROADCAST_GAME_STATE_SLEEP": 16,
@@ -32,6 +32,7 @@
     },
     "interpolation": {
         "REPLAY": true,
-        "REPLAY_TICK_DELTA_THRESHOLD": 5
+        "REPLAY_TICK_DELTA_THRESHOLD": 5,
+        "CLIENT_GAME_STATE_BUFFER_SIZE": 3
     }
 }
--- a/resources/configs/game_settings.json
+++ b/resources/configs/game_settings.json
@@ -25,7 +25,7 @@
     },
     "networking": {
         "SERVER_UNRELIABLE_RECV_TIMEOUT_SEC": 1,
-        "SERVER_TCP_PORT": 4845,
+        "SERVER_TCP_PORT": 4844,
         "SERVER_STATE_UDP_PORT": 4845,
         "SERVER_INPUT_UDP_PORT": 4846,
         "LOCALHOST_IP": "127.0.0.1"
--- a/resources/configs/game_settings.json
+++ b/resources/configs/game_settings.json
@@ -9,8 +9,11 @@
         "GAME_BOUNDS_RADIUS": 600.0,
         "TOTAL_RESOURCE_REQUIREMENTS": 20
     },
+    "core": {
+        "USE_REMAINDER_DELTA_TIME": false
+    },
     "physics": {
-        "MIN_TIME_STEP_SEC": 0.016
+        "MIN_TIME_STEP_SEC": 0.008
     },
     "threads": {
         "CLIENT_FETCH_PLAYER_ID_SLEEP": 16,
@@ -22,7 +25,7 @@
     },
     "networking": {
         "SERVER_UNRELIABLE_RECV_TIMEOUT_SEC": 1,
-        "SERVER_TCP_PORT": 4844,
+        "SERVER_TCP_PORT": 4845,
         "SERVER_STATE_UDP_PORT": 4845,
         "SERVER_INPUT_UDP_PORT": 4846,
         "LOCALHOST_IP": "127.0.0.1"
--- a/resources/configs/game_settings.json
+++ b/resources/configs/game_settings.json
@@ -28,11 +28,11 @@
         "SERVER_TCP_PORT": 4844,
         "SERVER_STATE_UDP_PORT": 4845,
         "SERVER_INPUT_UDP_PORT": 4846,
-        "LOCALHOST_IP": "127.0.0.1"
+        "LOCALHOST_IP": "127.0.0.1",
+        "INPUT_TICK_DELTA_THRESHOLD": 25
     },
     "interpolation": {
-        "REPLAY": true,
-        "REPLAY_TICK_DELTA_THRESHOLD": 5,
-        "CLIENT_GAME_STATE_BUFFER_SIZE": 3
+        "CLIENT_GAME_STATE_BUFFER_SIZE": 3,
+        "BEHIND_TICK_DELTA_THRESHOLD": 10
     }
 }
--- a/src/client/systems/ClientGameController.cpp
+++ b/src/client/systems/ClientGameController.cpp
@@ -121,6 +121,7 @@ void ClientGameController::update(const InputDef::PlayerInputs& pi, sf::Int32 de
     m_tickDeltaMetric.update();
     m_behindGameStateMetric.update();
     m_aheadGameStateMetric.update();
+    m_noGameStateMetric.update();
     m_interpolateGameStateMetric.update();
     m_stallGameStateMetric.update();
 }
@@ -133,6 +134,7 @@ void ClientGameController::postUpdate(sf::Int32 update_time) {
         .tick_delta_average = m_tickDeltaMetric.getAverage(),
         .behind_game_state_rate = m_behindGameStateMetric.getRate(sf::seconds(1)),
         .ahead_game_state_rate = m_aheadGameStateMetric.getRate(sf::seconds(1)),
+        .no_game_state_rate = m_noGameStateMetric.getRate(sf::seconds(1)),
         .interpolate_distance_average = m_interpolateGameStateMetric.getAverage(),
         .stall_distance_average = m_stallGameStateMetric.getAverage(),
         .resources = m_gameObjectController->getPlayerResources(m_playerId),
@@ -167,7 +169,7 @@ void ClientGameController::updateGameState(const InputDef::PlayerInputs& pi, sf:
         ahead(client_tick, smallest_server_tick, game_state_buffer);
         m_aheadGameStateMetric.pushCount();
     }
-    m_tickDeltaMetric.pushCount(smallest_server_tick - client_tick);
+    m_tickDeltaMetric.pushCount(std::abs(static_cast<int>(smallest_server_tick - client_tick)));
 }
 
 /**
--- a/src/client/systems/ClientGameController.cpp
+++ b/src/client/systems/ClientGameController.cpp
@@ -145,6 +145,7 @@ void ClientGameController::postUpdate(sf::Int32 update_time) {
         .game_status = m_gameStateCore.status,
         .winner_player_id = m_gameStateCore.winner_player_id,
         .tick = getTick(),
+        .tick_duration_ms = GAME_SETTINGS.MIN_TIME_STEP_SEC * 1000,
     };
     m_renderingController->postUpdate(update_time, ui_data);
 }
@@ -180,6 +181,7 @@ void ClientGameController::behind(uint32_t client_tick, uint32_t smallest_server
     uint32_t next_tick;
     if (smallest_server_tick - client_tick > MAX_BEHIND_TICK_DELTA) {
         next_tick = smallest_server_tick;
+
     } else {
         next_tick = client_tick + 1;
     }
@@ -222,10 +224,16 @@ void ClientGameController::interpolateGameState(uint32_t start_tick, uint32_t to
         throw std::runtime_error("Can't interpolate to the past.");
     }
 
-    float a = static_cast<float>(to_tick - start_tick) / static_cast<float>(end_tick - start_tick);
-    m_gameStateCore = game_state.core;
-    sf::Int32 delta_ms = MIN_TIME_STEP * (to_tick - start_tick);
-    m_gameObjectController->interpolateGameStateObject(game_state.object, a, delta_ms);
+    if (to_tick == end_tick) {
+        m_gameObjectController->applyGameStateObject(game_state.object);
+        m_gameStateCore = game_state.core;
+    } else {
+        float a =
+            static_cast<float>(to_tick - start_tick) / static_cast<float>(end_tick - start_tick);
+        m_gameStateCore = game_state.core;
+        sf::Int32 delta_ms = MIN_TIME_STEP * (to_tick - start_tick);
+        m_gameObjectController->interpolateGameStateObject(game_state.object, a, delta_ms);
+    }
     applyGameStateEvents(game_state.event);
     setTick(to_tick);
     m_interpolateGameStateMetric.pushCount(end_tick - to_tick);
--- a/src/client/systems/ClientGameController.cpp
+++ b/src/client/systems/ClientGameController.cpp
@@ -1,14 +1,19 @@
+#include <cmath>
 #include <iostream>
 #include <numeric>
 #include <thread>
 
 #include <SFML/Graphics.hpp>
 
+#include "../../common/events/CollisionEvent.hpp"
+#include "../../common/events/EventController.hpp"
 #include "../../common/physics/VectorUtil.hpp"
 #include "../../common/util/InputDef.hpp"
 #include "../../common/util/StateDef.hpp"
 #include "../../common/util/game_settings.hpp"
 #include "ClientGameController.hpp"
+
+const uint MAX_BEHIND_TICK_DELTA = 10;
 
 ClientGameController::ClientGameController(bool is_headless, bool is_bot, BotStrategy strategy,
                                            const std::string& server_ip, uint32_t server_tcp_port,
@@ -74,8 +79,11 @@ void ClientGameController::preUpdate() {
     switch (m_gameStateCore.status) {
         case GameStatus::not_started: {
             // Keep fetching game state to check if game status changed from not_started
-            GameState game_state = m_networkingClient->getLatestGameState();
-            m_gameStateCore = game_state.core;
+            std::map<uint32_t, GameState> game_state_buffer =
+                m_networkingClient->getGameStateBuffer();
+            if (game_state_buffer.size() == GAME_SETTINGS.CLIENT_GAME_STATE_BUFFER_SIZE) {
+                m_gameStateCore = game_state_buffer.begin()->second.core;
+            }
             break;
         }
         case GameStatus::playing: {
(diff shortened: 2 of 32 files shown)


src/client/systems/NetworkingClient.cpp
```diff
@@ -263,14 +259,33 @@ void NetworkingClient::unreliableRecv() {
...
         }
 
         std::this_thread::sleep_for(GAME_SETTINGS.CLIENT_UNRELIABLE_RECV_SLEEP);
     }
 }
 
+void NetworkingClient::updateGameStateBuffer(GameState& game_state) {
+    std::lock_guard<std::mutex> m_gameStateBuffer_guard(m_gameStateBuffer_lock);
+    std::lock_guard<std::mutex> m_serverGameStateMetric_guard(m_serverGameStateMetric_lock);
+    m_gameStateBuffer_t[game_state.core.tick] = game_state;
```
copacetic, line 272: im guessing a buffer overflow is impossible here since I believe this is a circular buffer iirc

src/client/systems/NetworkingClient.cpp
```diff
@@ -263,14 +259,33 @@ void NetworkingClient::unreliableRecv() {
...
+    if (m_gameStateBuffer_t.size() > GAME_SETTINGS.CLIENT_GAME_STATE_BUFFER_SIZE) {
+        if (m_gameStateBuffer_t.begin()->first != game_state.core.tick) {
```
copacetic, line 274: I dont understand this code, it's comparing the first entry in the game state buffer's tick to the the update's tick. if they aren't equal it updates the serverGameStateMetric. I guess im confused what serverGameStateMetric is supposed to be without grepping around. might be worth adding a comment explaining what's going on

copacetic: I dont get why it's erasing the first entry just because it doesnt equal the game state update we've received

copacetic: ie I thought the behavior would be that the smallest tick would be removed from the game state buffer *if* it was smaller than the incoming game state update's tick. if it's larger, than I thought the update would be discarded.

src/common/objects/CircleRigidBody.cpp
```diff
@@ -72,4 +72,18 @@ sf::Vector2f CircleRigidBody::getCenter() const {
     sf::Vector2f position = getPosition();
     float radius = getRadius();
     return sf::Vector2f(position.x + radius, position.y + radius);
-}
\ No newline at end of file
+}
+
+void CircleRigidBody::hermiteInterpolation(sf::Vector2f position, sf::Vector2f velocity, float a,
```
copacetic, line 77: it would be cool if you could add a summary here explaining what it's doing as a comment

src/common/objects/CircleRigidBody.hpp
```diff
@@ -39,6 +39,10 @@ class CircleRigidBody : public GameObject {
         m_position = position;
     }
 
+    void interpolatePosition(const sf::Vector2f position, float a) {
+        m_position = VectorUtil::lerp(m_position, position, a);
```
copacetic, line 43: should this not be using hermite interpolation? why not?

src/common/objects/CircleRigidBody.hpp
```diff
@@ -50,10 +54,20 @@ class CircleRigidBody : public GameObject {
     sf::Vector2f getTargetVelocity() const {
         return m_targetVelocity;
     }
+
     void setVelocity(sf::Vector2f velocity);
+
+    void interpolateVelocity(sf::Vector2f velocity, float a) {
+        m_position = VectorUtil::lerp(m_velocity, velocity, a);
```
copacetic, line 61: why no hermite

src/server/systems/NetworkingServer.hpp
```diff
@@ -23,7 +23,7 @@
 #include "../../common/util/StateDef.hpp"
 
 class NetworkingServer {
-    const uint32_t CMD_DRIFT_THRESHOLD = 7;
+    const uint32_t CMD_DRIFT_THRESHOLD = 1000000;
```
copacetic, line 26: is this still used? seems like it should be deleted since we have `GAME_SETTINGS.INPUT_TICK_DELTA_THRESHOLD` now

src/server/systems/ServerGameController.cpp
```diff
@@ -83,9 +90,19 @@ void ServerGameController::postUpdate(sf::Int32 update_time) {
...
     };
     m_networkingServer->setState(game_state);
+    m_gameStateEvent.collisions.clear();
+}
+
+void ServerGameController::handleCollisionEvent(std::shared_ptr<Event> event) {
+    std::shared_ptr<CollisionEvent> collision_event =
+        std::dynamic_pointer_cast<CollisionEvent>(event);
+
+    // save collision data
```
copacetic, line 104: is there no need to take a lock here?

src/server/systems/ServerGameController.cpp
```diff
@@ -83,9 +90,19 @@ void ServerGameController::postUpdate(sf::Int32 update_time) {
 void ServerGameController::setNetworkState() {
     GameState game_state = {
         .object = m_gameObjectController->getGameStateObject(),
+        .event = m_gameStateEvent,
         .core = m_gameStateCore,
     };
     m_networkingServer->setState(game_state);
+    m_gameStateEvent.collisions.clear();
```
copacetic, line 97: is there no lock needed here?

src/server/systems/ServerGameController.hpp
```diff
@@ -32,10 +35,12 @@ class ServerGameController : public GameController {
 
     // Methods
     void setNetworkState();
+    void addEventListeners();
 
     // Variables
     std::unique_ptr<NetworkingServer> m_networkingServer;
     TerminalRenderingController m_terminalRenderingController;
+    GameStateEvent m_gameStateEvent;
```
copacetic, line 43: What are the default values for field in GameStateEvents? Wondering if we need to 0 it out after creation to avoid random values

copacetic (commented):
very exciting change, the culmination of many weeks of reading and thinking :)

resources/configs/game_settings.json
```diff
@@ -9,12 +9,15 @@
         "GAME_BOUNDS_RADIUS": 600.0,
         "TOTAL_RESOURCE_REQUIREMENTS": 20
     },
+    "core": {
+        "USE_REMAINDER_DELTA_TIME": true
```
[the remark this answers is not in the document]
SourenP: This helps the server and client stay in sync, so I think it's better if it's on.
Now that the MIN_TIME_STEP_SEC is smaller (8ms), this should cause less jitter.

resources/configs/game_settings.json
```diff
@@ -25,10 +28,11 @@
...
         "SERVER_INPUT_UDP_PORT": 4846,
-        "LOCALHOST_IP": "127.0.0.1"
+        "LOCALHOST_IP": "127.0.0.1",
+        "INPUT_TICK_DELTA_THRESHOLD": 25
     },
     "interpolation": {
-        "REPLAY": true,
-        "REPLAY_TICK_DELTA_THRESHOLD": 5
+        "CLIENT_GAME_STATE_BUFFER_SIZE": 3,
+        "BEHIND_TICK_DELTA_THRESHOLD": 10
```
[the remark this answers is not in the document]
SourenP: Yes it's possible, and we handle it by stalling

src/client/systems/ClientGameController.cpp
```diff
@@ -132,73 +143,106 @@ void ClientGameController::postUpdate(sf::Int32 update_time) {
...
-        }
+    std::map<uint32_t, GameState> game_state_buffer = m_networkingClient->getGameStateBuffer();
 
-        m_largestAppliedGameStateTick = m_newGameStateTick;
-        m_tickDeltaMetric.pushCount(tick_delta);
-        m_networkUpdateMetric.pushCount();
-    } else if (m_newGameStateTick < m_largestAppliedGameStateTick) {
-        std::cout << "Recieved stale game state: " << m_newGameStateTick << "<"
-                  << m_largestAppliedGameStateTick << std::endl;
+    if (game_state_buffer.empty()) {
```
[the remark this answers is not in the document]
SourenP: Yeah it's possible. 
Going to add a debug line that shows you how many game states there are in the buffer and the tick delta of each of them

src/common/objects/CircleRigidBody.hpp
SourenP: hermite interpolation is only used for interpolating the position of CircleRigidBody given start pos & vel and end pos & vel

src/client/systems/ClientGameController.cpp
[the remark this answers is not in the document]
SourenP: I will also add a rate for how often this happens.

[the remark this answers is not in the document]
SourenP: Showing each states tick delta was too noisy, just added the latter.

src/client/systems/ClientGameController.cpp
```diff
@@ -132,73 +143,106 @@ void ClientGameController::postUpdate(sf::Int32 update_time) {
...
+    const GameState& smallest_server_game_state = game_state_buffer.begin()->second;
+
+    if (client_tick < smallest_server_tick) {
+        behind(client_tick, smallest_server_tick, smallest_server_game_state);
+        m_behindGameStateMetric.pushCount();
+    } else {
+        ahead(client_tick, smallest_server_tick, game_state_buffer);
+        m_aheadGameStateMetric.pushCount();
+    }
+    m_tickDeltaMetric.pushCount(smallest_server_tick - client_tick);
```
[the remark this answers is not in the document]
SourenP: good idea

src/client/systems/ClientGameController.cpp
```diff
@@ -132,73 +143,106 @@ void ClientGameController::postUpdate(sf::Int32 update_time) {
...
-    int tick_delta = client_tick - server_tick;
+/**
+ * If client is too far behind smallest server game state in the buffer, jump to server
+ * game state, otherwise interpolate the game state of the next tick.
+ */
+void ClientGameController::behind(uint32_t client_tick, uint32_t smallest_server_tick,
+                                  const GameState& smallest_server_game_state) {
+    uint32_t next_tick;
+    if (smallest_server_tick - client_tick > GAME_SETTINGS.BEHIND_TICK_DELTA_THRESHOLD) {
+        next_tick = smallest_server_tick - GAME_SETTINGS.BEHIND_TICK_DELTA_THRESHOLD;
```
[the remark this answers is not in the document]
SourenP: Not sure what you mean. Maybe you misread the code? 

If client tick is farther than the threshold _amount_ then we'll jump it ahead to the edge.

So if
- `server_tick` is 10
- `tick_delta_thresh` is 4
- `client_tick` is 2

If `server_tick` - `client_tick` (8) > `tick_delta_thresh` (4)
Then make the next client tick be `server_tick` - `tick_delta_thresh` (6)

src/client/systems/ClientGameController.cpp
```diff
@@ -132,73 +143,106 @@ void ClientGameController::postUpdate(sf::Int32 update_time) {
...
-                m_playerId, client_input_and_tick.ri, client_input_and_tick.ui));
-            GameController::computeGameState(pi, GameController::MIN_TIME_STEP);
-        } else {
-            // If we don't have input for this tick pass in empty inputs
-            auto pi = InputDef::PlayerInputs();
-            GameController::computeGameState(pi, GameController::MIN_TIME_STEP);
+/**
+ * If client is ahead of the smallest server game state in the buffer, stay at the same tick but
+ * interpolate to the next available game state in the buffer. If there are none available just keep
+ * the game state the same. Note that the tick stays the same in this case so that we fall back
```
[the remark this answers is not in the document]
SourenP: yes

src/client/systems/ClientGameController.cpp
```diff
@@ -132,73 +143,106 @@ void ClientGameController::postUpdate(sf::Int32 update_time) {
...
-    // future replays because we never apply stale game states.
-    for (auto it = m_tickToInput.begin(); it != m_tickToInput.end();) {
-        if (it->first < server_tick) {
-            m_tickToInput.erase(it++);
-        } else {
-            ++it;
-        }
+    // If we reach here then we don't have any game states to interpolate to. Keep the same tick and
+    // game state.
+    m_stallGameStateMetric.pushCount(client_tick - smallest_server_tick);
```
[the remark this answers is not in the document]
SourenP: `m_aheadGameStateMetric` tracks how many times the game is stalled
`m_stallGameStateMetric` tracks how far ahead we were when we had to stall

src/client/systems/ClientGameController.cpp
```diff
@@ -132,73 +143,106 @@ void ClientGameController::postUpdate(sf::Int32 update_time) {
...
+    return;
```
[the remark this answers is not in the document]
SourenP: Yes
Even when we stall, we use the next game state we have to interpolate the same tick we're currently on (in case our prediction from before that got us in the state we are was wrong). 
We only want to do this for the next game state. If we don't return then we'll keep doing this for every game state head of us in the buffer.

now that i think of it tho, maybe we should using the _latest_ game state we have, that's the most accurate.....

src/client/systems/ClientGameController.cpp
```diff
@@ -132,73 +143,106 @@ void ClientGameController::postUpdate(sf::Int32 update_time) {
...
+}
+
+void ClientGameController::interpolateGameState(uint32_t start_tick, uint32_t to_tick,
```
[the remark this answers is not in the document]
SourenP: it's in the hpp file
I made the comment more descriptive

src/client/systems/ClientGameController.cpp
```diff
@@ -132,73 +143,106 @@ void ClientGameController::postUpdate(sf::Int32 update_time) {
...
+        m_gameStateCore = game_state.core;
+    } else {
+        float a =
+            static_cast<float>(to_tick - start_tick) / static_cast<float>(end_tick - start_tick);
+        m_gameStateCore = game_state.core;
+        sf::Int32 delta_ms = MIN_TIME_STEP * (to_tick - start_tick);
+        m_gameObjectController->interpolateGameStateObject(game_state.object, a, delta_ms);
     }
+    applyGameStateEvents(game_state.event);
+    setTick(to_tick);
```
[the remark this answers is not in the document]
SourenP: Yeah

src/client/systems/ClientGameController.hpp
```diff
@@ -54,45 +55,63 @@ class ClientGameController : public GameController {
 
     void draw();
 
-    /*
-     * Apply the game state update from the server (rewind, because it is likely the state of an old
-     * tick) and then interpolate up to the current tick by applying local inputs for each tick
-     * (replay).
+    /**
+     * Update game state by stalling, interpolating and/or extrapolating.
```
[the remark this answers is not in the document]
SourenP: Left it in there so that we consider extrapolating in the future 🙃
hence the _or_

src/client/systems/NetworkingClient.cpp
SourenP: it's a map not a circular buffer

src/common/objects/CircleRigidBody.hpp
SourenP: This is linear interpolation
I should update the name

We use hermite but i left this in there in case we ever want to do linear interpolation

src/server/systems/NetworkingServer.hpp
SourenP: oops i thought i left it out

src/server/systems/ServerGameController.cpp
SourenP: No, this is just date passed in to this function

src/server/systems/ServerGameController.cpp
SourenP: no, the lock happens in 
`m_networkingServer->setState(game_state);`

m_gameStateEvent is just a local variable only used by this class

src/server/systems/ServerGameController.hpp
SourenP:
```
struct GameStateEvent {
    std::vector<Collision> collisions;
};
```

A vectors default value is just an empty list

src/client/systems/NetworkingClient.cpp
SourenP: This code got a bit more complicated because I wanted to track when we put something new into the buffer for `m_serverGameStateMetric`

But essentially what's happening is 
272: we insert the new game state
273-277: if the exceeded our size then we delete the smallest member, if the smallest member isn't the think we just entered then it's a new game state (that won't be removed) so update the count.

src/common/objects/CircleRigidBody.cpp
SourenP: it's hard to explain and I dont really understand it well
i expect the reader to just google "hermite interpolation" to figure it out

SourenP:
@copacetic I'll leave it to you to resolve comments

src/client/systems/ClientGameController.cpp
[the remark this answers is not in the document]
copacetic: nice good change

src/client/systems/ClientGameController.cpp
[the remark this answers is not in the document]
copacetic: I guess Im wondering: why do we subtract the threshold value? what are we achieving by doing this? (what's the point of this)

src/client/systems/ClientGameController.cpp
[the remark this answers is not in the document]
copacetic: i thought that was extrapolation, is it just that it's out of scope for now or does extrap not actually fix this?

src/client/systems/ClientGameController.cpp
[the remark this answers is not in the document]
copacetic: I guess I mean this is the last statement in the function... i thought functions returned null by default if you didn't explicitly put a statement, so im wondering if this return statement is unnecessary

src/client/systems/ClientGameController.hpp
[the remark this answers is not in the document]
copacetic: ehhhh Id remove it it's purely confusing and adds 0 benefit

